PR #27, opened 2020-03-17
docs(firestore): on_snapshot document changes

### LINKED ISSUE #26: inconsistent type annotations for Document.on_snapshot
Looking here (which mirrors the API docs):
https://github.com/googleapis/python-firestore/blob/de3aca0e78b68f66eb76bc679c6e95b0746ad590/google/cloud/firestore_v1/document.py#L503-L504

The claim is that the callback invoked by `on_snapshot` should have type signature `Callable[[:class:`~google.cloud.firestore.document.DocumentSnapshot`], NoneType]`. however, just a few lines lower, in the example, it looks like the callback receives three arguments: https://github.com/googleapis/python-firestore/blob/de3aca0e78b68f66eb76bc679c6e95b0746ad590/google/cloud/firestore_v1/document.py#L515

i tried to understand what's going on by reading the code, and didn't make too much progress. we invoke the callback here: https://github.com/googleapis/python-firestore/blob/de3aca0e78b68f66eb76bc679c6e95b0746ad590/google/cloud/firestore_v1/watch.py#L569-L573

it looks like there are indeed three arguments passed. `appliedChanges` appears to be a `List[DocumentChange]`, and the final argument seems to be a `datetime.datetime` instance. but the first argument... is the result of calling `keys()` on a `document_tree` (here: https://github.com/googleapis/python-firestore/blob/de3aca0e78b68f66eb76bc679c6e95b0746ad590/google/cloud/firestore_v1/watch.py#L567) and so maybe is a `List[str]` ? the `self.doc_tree` is an instance of `WatchDocTree` and it's `keys()` method returns the underlying dict's keys. are those keys actually `DocumentSnapshot` instances? it maybe sort of seems like it, after tracing through the code a bunch. i'm *guessing* that the correct type signature for the callback is:

```
Callable[List[DocumentSnapshot], List[DocumentChange], datetime.datime], None]
```

but it would be nice if this was confirmed, and the documentation updated to reflect reality.

HemangChothani:
Fixes #26


BenWhitehead:
@crwilcox or @tseaver Can either of you review this? I'm not familiar with pythons type annotations.

crwilcox:
@HemangChothani can you update this branch?

Also, not sure why this change is being made? It seems to edit actual types of functions, not just the annotations.

crwilcox:
https://github.com/googleapis/python-firestore/pull/79 opened with me merging master. Thanks!

PR #40, opened 2020-05-06
feat: Create CODEOWNERS

czahedi:
Adding owner team


czahedi:
Trying a comment after changing the title to fix the conventional commits thing

PR #57, opened 2020-06-04
feat: support limit to last feature

IlyaFaer:
Pushing an implementation with de-deprecation of the `get()` method (as @BenWhitehead mentioned).

### DIFF
--- a/google/cloud/firestore_v1/collection.py
+++ b/google/cloud/firestore_v1/collection.py
@@ -272,6 +272,24 @@ class CollectionReference(object):
         query = query_mod.Query(self)
         return query.limit(count)
 
+    def limit_to_last(self, count):
+        """Create a limited query with this collection as parent.
+
+        See
+        :meth:`~google.cloud.firestore_v1.query.Query.limit` for
+        more information on this method.
+
+        Args:
+            count (int): Maximum number of documents to return that match
+                the query.
+
+        Returns:
+            :class:`~google.cloud.firestore_v1.query.Query`:
+            A limited query.
+        """
+        query = query_mod.Query(self)
+        return query.limit_to_last(count)
+
     def offset(self, num_to_skip):
         """Skip to an offset in a query with this collection as parent.
 
--- a/google/cloud/firestore_v1/collection.py
+++ b/google/cloud/firestore_v1/collection.py
@@ -14,7 +14,6 @@
 
 """Classes for representing collections for the Google Cloud Firestore API."""
 import random
-import warnings
 
 import six
 
@@ -273,19 +272,19 @@ class CollectionReference(object):
         return query.limit(count)
 
     def limit_to_last(self, count):
-        """Create a limited query with this collection as parent.
+        """Create a limited to last query with this collection as parent.
 
         See
         :meth:`~google.cloud.firestore_v1.query.Query.limit` for
         more information on this method.
 
         Args:
-            count (int): Maximum number of documents to return that match
-                the query.
+            count (int): Maximum number of documents to return that
+                match the query.
 
         Returns:
             :class:`~google.cloud.firestore_v1.query.Query`:
-            A limited query.
+            A limited to last query.
         """
         query = query_mod.Query(self)
         return query.limit_to_last(count)
@@ -393,13 +392,25 @@ class CollectionReference(object):
         return query.end_at(document_fields)
 
     def get(self, transaction=None):
-        """Deprecated alias for :meth:`stream`."""
-        warnings.warn(
-            "'Collection.get' is deprecated:  please use 'Collection.stream' instead.",
-            DeprecationWarning,
-            stacklevel=2,
-        )
-        return self.stream(transaction=transaction)
+        """Read the documents in this collection.
+
+        This sends a ``RunQuery`` RPC and returns a list of documents
+        returned in the stream of ``RunQueryResponse`` messages.
+
+        Args:
+            transaction
+                (Optional[:class:`~google.cloud.firestore_v1.transaction.Transaction`]):
+                An existing transaction that this query will run in.
+
+        If a ``transaction`` is used and it already has write operations
+        added, this method cannot be used (i.e. read-after-write is not
+        allowed).
+
+        Returns:
+            list: The documents in the collection that match this query.
+        """
+        query = query_mod.Query(self)
+        return query.get(transaction=transaction)
 
     def stream(self, transaction=None):
         """Read the documents in this collection.
--- a/google/cloud/firestore_v1/collection.py
+++ b/google/cloud/firestore_v1/collection.py
@@ -256,6 +256,11 @@ class CollectionReference(object):
     def limit(self, count):
         """Create a limited query with this collection as parent.
 
+        .. note::
+
+           `limit` and `limit_to_last` are mutually exclusive.
+           Setting `limit` will drop previously set `limit_to_last`.
+
         See
         :meth:`~google.cloud.firestore_v1.query.Query.limit` for
         more information on this method.
@@ -273,6 +278,11 @@ class CollectionReference(object):
 
     def limit_to_last(self, count):
         """Create a limited to last query with this collection as parent.
+
+        .. note::
+
+           `limit` and `limit_to_last` are mutually exclusive.
+           Setting `limit_to_last` will drop previously set `limit`.
 
         See
         :meth:`~google.cloud.firestore_v1.query.Query.limit_to_last`
--- a/google/cloud/firestore_v1/collection.py
+++ b/google/cloud/firestore_v1/collection.py
@@ -275,8 +275,8 @@ class CollectionReference(object):
         """Create a limited to last query with this collection as parent.
 
         See
-        :meth:`~google.cloud.firestore_v1.query.Query.limit` for
-        more information on this method.
+        :meth:`~google.cloud.firestore_v1.query.Query.limit_to_last`
+        for more information on this method.
 
         Args:
             count (int): Maximum number of documents to return that
@@ -407,7 +407,7 @@ class CollectionReference(object):
         allowed).
 
         Returns:
-            list: The documents in the collection that match this query.
+            list: The documents in this collection that match the query.
         """
         query = query_mod.Query(self)
         return query.get(transaction=transaction)
--- a/google/cloud/firestore_v1/query.py
+++ b/google/cloud/firestore_v1/query.py
@@ -351,6 +351,11 @@ class Query(object):
 
         If the current query already has a `limit` set, this will override it.
 
+        .. note::
+
+           `limit` and `limit_to_last` are mutually exclusive.
+           Setting `limit` will drop previously set `limit_to_last`.
+
         Args:
             count (int): Maximum number of documents to return that match
                 the query.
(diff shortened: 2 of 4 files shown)


google/cloud/firestore_v1/query.py
```diff
@@ -729,13 +761,37 @@ def _to_protobuf(self):
...
+            self._limit = self._limit_to_last
+            self._limit_to_last = None
+
+            # flip order statements
+            for order in self._orders:
+                order.direction = _enum_from_direction(
+                    self.DESCENDING
+                    if order.direction == self.ASCENDING
+                    else self.ASCENDING
+                )
```
IlyaFaer, line 791: Before sending a backend request we're transforming `_limit_to_last` to `_limit` and flipping order statements.

google/cloud/firestore_v1/query.py
```diff
@@ -729,13 +761,37 @@ def _to_protobuf(self):
...
+
+        result = list(self.stream(transaction=transaction))
+        return list(reversed(result))
```
IlyaFaer, line 794: Calling `list()` to stream data into `list`, then reversing it (after this comment we'll get an iterator) and converting back to `list`.

tests/unit/v1/test_query.py
```diff
@@ -1104,9 +1102,60 @@ def test_get_simple(self):
...
+        # Add a dummy response to the minimal fake GAPIC.
+        _, expected_prefix = parent._parent_info()
+        name = "{}/sleep".format(expected_prefix)
+        data = {"snooze": 10}
+        data2 = {"snooze": 20}
+
+        response_pb = _make_query_response(name=name, data=data)
+        response_pb2 = _make_query_response(name=name, data=data2)
+
+        firestore_api.run_query.return_value = iter([response_pb2, response_pb])
```
IlyaFaer, line 1128: Return responses in reversed order (as it's gonna be in real cases)

tests/unit/v1/test_query.py
```diff
@@ -1104,9 +1102,60 @@ def test_get_simple(self):
...
+        query = self._make_one(parent)
+        query = query.order_by(
+            u"snooze", direction=firestore.Query.DESCENDING
+        ).limit_to_last(2)
+        returned = query.get()
+
+        self.assertIsInstance(returned, list)
+        self.assertEqual(
+            query._orders[0].direction, _enum_from_direction(firestore.Query.ASCENDING),
+        )
```
IlyaFaer, line 1140: Check if order statements were reversed

tests/unit/v1/test_query.py
```diff
@@ -1104,9 +1102,60 @@ def test_get_simple(self):
...
+        self.assertEqual(len(returned), 2)
+
+        snapshot = returned[0]
+        self.assertEqual(snapshot.reference._path, ("dee", "sleep"))
+        self.assertEqual(snapshot.to_dict(), data)
+
+        snapshot2 = returned[1]
+        self.assertEqual(snapshot2.reference._path, ("dee", "sleep"))
+        self.assertEqual(snapshot2.to_dict(), data2)
```
IlyaFaer, line 1149: Check that data in the result `list` were reversed before return

IlyaFaer:
Should we add a system test for this feature?

BenWhitehead:
@crwilcox I'll defer to you on the python code change.

google/cloud/firestore_v1/collection.py
```diff
@@ -375,13 +392,25 @@ def end_at(self, document_fields):
...
 
     def get(self, transaction=None):
-        """Deprecated alias for :meth:`stream`."""
-        warnings.warn(
-            "'Collection.get' is deprecated:  please use 'Collection.stream' instead.",
-            DeprecationWarning,
-            stacklevel=2,
-        )
-        return self.stream(transaction=transaction)
+        """Read the documents in this collection.
```
crwilcox, line 395: I might be missing something, but is this de-deprecation related to limit to last?

google/cloud/firestore_v1/collection.py
```diff
@@ -272,6 +271,24 @@ def limit(self, count):
         query = query_mod.Query(self)
         return query.limit(count)
 
+    def limit_to_last(self, count):
```
crwilcox, line 274: is adding this to collection required? does this match other language implementations better?

google/cloud/firestore_v1/query.py
```diff
@@ -333,16 +339,17 @@ def order_by(self, field_path, direction=ASCENDING):
...
             end_at=self._end_at,
             all_descendants=self._all_descendants,
         )
 
     def limit(self, count):
-        """Limit a query to return a fixed number of results.
+        """Limit a query to return first `count` results.
 
-        If the current query already has a limit set, this will overwrite it.
+        If the current query already has a `limit` set, this will overwrite it.
```
crwilcox, line 352: should this be overwrite or override?

google/cloud/firestore_v1/query.py
```diff
@@ -333,16 +339,17 @@ def order_by(self, field_path, direction=ASCENDING):
...
+            limit_to_last=self._limit_to_last,
             offset=self._offset,
             start_at=self._start_at,
             end_at=self._end_at,
             all_descendants=self._all_descendants,
         )
 
     def limit(self, count):
-        """Limit a query to return a fixed number of results.
+        """Limit a query to return first `count` results.
```
crwilcox, line 350: I think this is missing an article. "the first `count` results". 

Also, is it the first, or that many results from the point of the query? For instance, if the query is set to skip (offset) the first 100 results, this might be misleading

google/cloud/firestore_v1/collection.py
IlyaFaer: There is a [test](https://github.com/googleapis/python-firestore/blob/b1c5987c606a14874b412e70f93015e161e278d6/tests/unit/v1/test_collection.py#L41-L49) which checks that `Query()` has the same (or less) methods that `CollectionReference()` does. So, as I understood that, the list of the supported methods should be the same for `Query()` and `CollectionReference()`, at least in our case.

google/cloud/firestore_v1/collection.py
IlyaFaer: Yes, we need to reverse result came from backend before returning it to user, so streaming is not our way, and `get()` method becomes the best choice. As for de-deprecating this method in `CollectionReference()`, see the other comment about methods matching:

> There is a test which checks that Query() has the same (or less) methods that CollectionReference() does. So, as I understood that, the list of the supported methods should be the same for Query() and CollectionReference(), at least in our case.

google/cloud/firestore_v1/query.py
IlyaFaer: Changed to `override`

google/cloud/firestore_v1/query.py
IlyaFaer: Hm-m, looks like I've missed the word "matching":

> Limit a query to return the last `count` matching results.

Does this make more sense?

crwilcox: It does, though still wondering if last and first make sense? it could actually be the middle `count` results.

For instance, if I offset by 10, limit to 10, but have 30 results, I actually am viewing results 10-20

google/cloud/firestore_v1/query.py
```diff
@@ -86,6 +85,8 @@ class Query(object):
             The "order by" entries to use in the query.
         limit (Optional[int]):
             The maximum number of documents the query is allowed to return.
+        limit_to_last (Optional[bool]):
+            Limit query results to last.
```
crwilcox, line 89: to last what?

google/cloud/firestore_v1/query.py
crwilcox: Perhaps `Limit a query to return the next `count` matching results.`?

google/cloud/firestore_v1/query.py
IlyaFaer: Oh, looks like I've lost the think just in a middle of an action. Rewritten it in Java's comment fashion

google/cloud/firestore_v1/query.py
```diff
@@ -333,16 +339,17 @@ def order_by(self, field_path, direction=ASCENDING):
...
+            limit_to_last=self._limit_to_last,
             offset=self._offset,
             start_at=self._start_at,
             end_at=self._end_at,
             all_descendants=self._all_descendants,
         )
 
     def limit(self, count):
-        """Limit a query to return a fixed number of results.
+        """Limit a query to return the next `count` matching results.
```
crwilcox, line 350: @BenWhitehead I had @IlyaFaer change this but after our chat I wanted you to double check

BenWhitehead: I've taken a look, and started a review of my own, with some recommendations.

BenWhitehead, line 350: Slight rephrasing to be more inline with how the query functions:

```suggestion
        """Limit a query to return at most `count` matching results.
```

google/cloud/firestore_v1/query.py
```diff
@@ -333,16 +339,17 @@ def order_by(self, field_path, direction=ASCENDING):
...
 
-        If the current query already has a limit set, this will overwrite it.
+        If the current query already has a `limit` set, this will override it.
```
BenWhitehead, line 352: Can we add a note here stating that `limit` and `limit_to_last` are mutually exclusive?

i.e. this: `query.limit_to_last(10).limit(5)` doesn't really make sense. The result will end up being the same as `query.limit(5)`

google/cloud/firestore_v1/query.py
```diff
@@ -359,6 +366,35 @@ def limit(self, count):
...
+            all_descendants=self._all_descendants,
+        )
+
+    def limit_to_last(self, count):
+        """Limit a query to return the last `count` matching results.
+
+        If the current query already has a `limit_to_last`
+        set, this will override it.
+
+        Args:
```
BenWhitehead, line 382: Can we add the note here about mutual exclusivity with `limit` and `limit_to_last` here too?

google/cloud/firestore_v1/query.py
```diff
@@ -729,13 +766,40 @@ def _to_protobuf(self):
...
+        added, this method cannot be used (i.e. read-after-write is not
+        allowed).
+
+        Returns:
+            list: The documents in the collection that match this query.
+        """
+        is_limited_to_last = self._limit_to_last
+
+        if self._limit_to_last:
+            # flip order statements
```
BenWhitehead, line 789: We probably want an expanded explanation comment here to illustrate why we're doing the flip.

Something like:
```suggestion
            # In order to fetch up to `self._limit` results from the end of the 
            # query flip the defined ordering on the query to start from the 
            # end, retrieving up to `self._limit` results from the backend.

```

google/cloud/firestore_v1/collection.py
```diff
@@ -272,6 +271,24 @@ def limit(self, count):
...
+        """Create a limited to last query with this collection as parent.
+
+        See
```
BenWhitehead, line 277: Can we add the note here about mutual exclusivity with `limit` and `limit_to_last` here too?

IlyaFaer: Done

google/cloud/firestore_v1/query.py
IlyaFaer: Done

google/cloud/firestore_v1/query.py
IlyaFaer: Done

crwilcox:
@IlyaFaer @BenWhitehead I think we may want this in before releasing 1.8.0

BenWhitehead:
Yeah, we definitely want this pre v2

crwilcox:
Thanks for the updated comments. Merging.

PR #58, opened 2020-06-05
fix(firestore): use specific naming convention

### LINKED ISSUE #56: Support DocumentSnapshot cursors
The Python client should add support for DocumentSnapshot-based cursors.

See https://github.com/googleapis/java-firestore/blob/master/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java#L853 or https://github.com/googleapis/nodejs-firestore/blob/master/dev/src/reference.ts#L1663

HemangChothani:
Towards #56

### DIFF
--- a/google/cloud/firestore_v1/query.py
+++ b/google/cloud/firestore_v1/query.py
@@ -392,7 +392,7 @@ class Query(object):
             all_descendants=self._all_descendants,
         )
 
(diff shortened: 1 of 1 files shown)


tseaver:
This PR needs to be ported to the new `master` branch (used to be `v2-staging`).

PR #59, opened 2020-06-17
fix: Python requires didn't exclude 3.4 to match classifier

crwilcox:
commit 48339ec removed the 3.4 classifier, but requires wasn't updated.

We could consider this breaking as it will prevent installs. We could also hold this until we drop 3.5, likely September.


crwilcox:
NOTE: seems this is more of a broad pattern. I think I plan to *not* merge this, instead batching this with the likely September major version release.

PR #61, opened 2020-06-22
feat: create async interface

rafilong:
Add async interface and testers.


google/cloud/firestore_v1/async_collection.py
```diff
@@ -0,0 +1,373 @@
...
+        Args:
+            field_paths (Iterable[str, ...]): An iterable of field paths
+                (``.``-delimited list of field names) to use as a projection
+                of document fields in the query results.
+
+        Returns:
+            :class:`~google.cloud.firestore_v1.async_query.AsyncQuery`:
+            A "projected" query.
+        """
+        query = async_query.AsyncQuery(self)
```
crwilcox, line 132: after looking at this, I think we may want to have a core class for collection with small modifications to allow async, since most of these things aren't async specific. Duplicated code seems likely to drift apart. wdyt @BenWhitehead

rafilong: SGTM! We can reduce bloat in the testers as well, should cut down line numbers significantly.

rafilong:
Pulled out base class refactor to PRs #66 #67 #68 #69 #70 to reduce the size of this PR.

rafilong:
Adding `do not merge` label - working on a `Transaction` refactor using multiple inheritance that should reduce clutter. PR coming soon.

rafilong:
Base class refactor now includes #81, which utilizes multiple inheritance to refactor `transaction.py`.

tests/unit/v1/async/test_async_query.py
```diff
@@ -0,0 +1,380 @@
...
+        self.assertEqual(query._field_filters, ())
+        self.assertEqual(query._orders, ())
+        self.assertIsNone(query._limit)
+        self.assertIsNone(query._offset)
+        self.assertIsNone(query._start_at)
+        self.assertIsNone(query._end_at)
+        self.assertFalse(query._all_descendants)
+
+    @pytest.mark.asyncio
+    async def test_get_simple(self):
```
BenWhitehead, line 48: What does `simple` mean in this test name? I don't know that it's the responsibility of this test to verify the code that is responsible for translating a query to the protobuf request. So I'm not sure I understand what simple is alluding to.

tests/unit/v1/async/test_async_transaction.py
```diff
@@ -0,0 +1,1055 @@
...
+
+        client = _make_client()
+        transaction = self._make_one(client)
+        transaction._id = b"not-none"
+
+        with self.assertRaises(ValueError) as exc_info:
+            await transaction._begin()
+
+        err_msg = _CANT_BEGIN.format(transaction._id)
+        self.assertEqual(exc_info.exception.args, (err_msg,))
```
BenWhitehead, line 126: This feels like a hollow assertion since we're constructing the expected result in the exact same why the internal code is. It may be better to actually write out the expected message to verify the internal creation of it does what is desired.

tests/unit/v1/async/test_async_transaction.py
```diff
@@ -0,0 +1,1055 @@
...
+
+        # Call function and check result.
+        txn_id = b"the-world\x00"
+        commit_response = await self._call_fut(client, mock.sentinel.write_pbs, txn_id)
+        self.assertIs(commit_response, mock.sentinel.commit_response)
+
+        # Verify mocks used.
+        self.assertEqual(_sleep.call_count, 2)
+        _sleep.assert_any_call(1.0)
+        _sleep.assert_any_call(2.0)
```
BenWhitehead, line 869: Is `_sleep` here saying "wait for 1 event to finish, then wait for the 2nd event to finish" or "wait for 1 event to finish, then wait for 2 more events to finish"?

google/cloud/firestore_v1/async_client.py
```diff
@@ -0,0 +1,288 @@
+# Copyright 202 Google LLC All rights reserved.
```
BenWhitehead: `s/202/2020/`

rafilong:
@BenWhitehead my thought is to leave system tests until the async microgen client is intregrated, as that way we can test for async functionality. As is all of the classes are stubbed out for async, but the actual networked calls are still blocking.

Also, just to note, most of the current async unit tests are ports from the sync test cases. In that regard, your comments are applicable to both of them. Would it make sense to fix both the async and sync unit tests in this PR, or should I make a new PR to do that?

tests/unit/v1/async/test_async_client.py
```diff
@@ -0,0 +1,462 @@
...
+
+    def test_collection_group(self):
+        client = self._make_default_one()
+        query = client.collection_group("collectionId").where("foo", "==", u"bar")
+
+        assert query._all_descendants
+        assert query._field_filters[0].field.field_path == "foo"
+        assert query._field_filters[0].value.string_value == u"bar"
+        assert query._field_filters[0].op == query._field_filters[0].Operator.EQUAL
+        assert query._parent.id == "collectionId"
```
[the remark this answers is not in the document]
rafilong: fixed!

tests/unit/v1/async/__init__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2017 Google LLC
```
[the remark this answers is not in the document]
rafilong: fixed!

crwilcox:
Possibly worth noting. This PR does not contain integration of async generated bits. While the surface is async, the underlying code isn't yet.

BenWhitehead:
Sounds good, apologies for not understanding the async integration with generated client hasn't already taken place. I think it's fine to leave adding integration tests until that time.

Regarding changing both the async and sync unit tests I don't think that is probably necessary as long as there are integration tests that will come. The thing that stood out to me is that there weren't integration tests added in this PR.

rafilong:
Sorry I wasn't very clear about integration of async, but the changes to the async and sync tests I was referring to was with regards to the fixes you suggested for the async unit tests. Would it be best to integrate these changes into this PR (like I did with the fixing assert syntax), or bundle them in another PR?

BenWhitehead:
Up to you, either is fine with me.

tests/unit/v1/async/test_async_document.py
```diff
@@ -0,0 +1,511 @@
...
+        firestore_api = mock.Mock(spec=["commit"])
+        firestore_api.commit.return_value = self._make_commit_repsonse()
+
+        # Attach the fake GAPIC to a real client.
+        client = _make_client("db-dee-bee")
+        client._firestore_api_internal = firestore_api
+
+        # Actually make a document and call create().
+        document = self._make_one("User", "Interface", client=client)
+        document_data = {"And": 500, "Now": b"\xba\xaa\xaa \xba\xaa\xaa"}
```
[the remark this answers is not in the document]
rafilong: Yep, I believe so

tests/unit/v1/async/test_async_query.py
rafilong: Similar to `test_stream_simple`, its referring to a simple usecase of `get`. I think this is confusing because there are no other `get` test cases, as the `get` implementation wraps `stream`. Maybe it would be better if `get` was tested with mocks and we scrapped this test altogether?

tests/unit/v1/async/test_async_transaction.py
rafilong: The `_sleep` mock assert check here is saying "make sure there were two calls to sleep, where one call to sleep was 1 second and the another call to sleep was 2 seconds". `_sleep` is called in `_commit_with_retry` with multiplicatively larger intervals. The test case is checking the functionality of `_commit_with_retry` when the commit fails the first two times and succeeds the third, in which case the expected sleep is 1 second and then 2 seconds.

tests/unit/v1/async/test_async_query.py
BenWhitehead: I think that's probably fair since `get` just calls `stream`.

tests/unit/v1/async/test_async_transaction.py
BenWhitehead: Ah, okay so it's ensuring there is backoff.

Great, can we add a comment stating as much?

tests/unit/v1/async/test_async_query.py
rafilong: Sounds good! I'll make that change.

tests/unit/v1/async/test_async_transaction.py
rafilong: Sure!

tests/unit/v1/async/test_async_query.py
rafilong: https://github.com/googleapis/python-firestore/issues/101

tests/unit/v1/async/test_async_collection.py
```diff
@@ -0,0 +1,363 @@
...
+            collection_id1, document_id, collection_id2, client=client
+        )
+        self.assertIs(collection._client, client)
+        expected_path = (collection_id1, document_id, collection_id2)
+        self.assertEqual(collection._path, expected_path)
+
+    def test_constructor_invalid_path(self):
+        with self.assertRaises(ValueError):
+            self._make_one()
+        with self.assertRaises(ValueError):
```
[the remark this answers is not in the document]
rafilong: #102

tests/unit/v1/async/test_async_document.py
```diff
@@ -0,0 +1,511 @@
...
+            collection_id1, document_id1, collection_id2, document_id2, client=client
+        )
+        self.assertIs(document._client, client)
+        expected_path = "/".join(
+            (collection_id1, document_id1, collection_id2, document_id2)
+        )
+        self.assertEqual(document.path, expected_path)
+
+    def test_constructor_invalid_path(self):
+        with self.assertRaises(ValueError):
```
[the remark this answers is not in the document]
rafilong: #102

tests/unit/v1/async/test_async_transaction.py
rafilong: That's a good point. If the construction of the error message was marginally more complex I would agree with you hands down, but it feels as though this is acceptable as it is very close to comparing a defined constant string. @crwilcox do you have any strong opinions about this?

tests/unit/v1/async/test_async_batch.py
```diff
@@ -0,0 +1,158 @@
...
+        with self.assertRaises(RuntimeError):
+            async with batch as ctx_mgr:
+                ctx_mgr.create(document1, {"ten": 10, "buck": u"ets"})
+                ctx_mgr.delete(document2)
+                raise RuntimeError("testing")
+
+        self.assertIsNone(batch.write_results)
+        self.assertIsNone(batch.commit_time)
+        # batch still has its changes
+        self.assertEqual(len(batch._write_pbs), 2)
```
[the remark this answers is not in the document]
rafilong: I'm not entirely sure about this, but from what I understand is that when an error is raised inside the context, `_aexit_` / `_exit_` are not called. In this case it means `commit` is not run, so the `_write_pbs` are not set to empty.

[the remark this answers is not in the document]
rafilong: @BenWhitehead I expanded on the existing comment; do you think this is sufficient, or is more detail needed?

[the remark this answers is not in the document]
crwilcox: I think this is due to an error while trying to make a batch commit results in the batch not being unset, so you could attempt  to correct, try again?

tests/unit/v1/async/test_async_transaction.py
crwilcox: A lot of the exception checking is fragile. We already encountered an issue with this in the microgen conversion. Basically we can use the `assertRaisesRegex` instead but it isn't much better. We could mock the var so we could see it was called in the impl properly? But again not sure it is better?

TL;DR - ack

tests/unit/v1/async/test_async_batch.py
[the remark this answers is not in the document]
rafilong: Yep, exactly. Only in the case that the context is exited successfully is a commit attempted, and only on success are the staged `_write_pbs` set to empty.

crwilcox:
It seems any issues are either tracked in issues, or have been addressed. I think this is good to merge.

tests/unit/v1/async/test_async_batch.py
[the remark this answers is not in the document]
rafilong: Created issue #107 and added rationale to comment.

Issue #65, opened 2020-06-25
Refactor to generalize classes


rafilong:
Refactor classes with the intent of adding async support. Refactor will enable future async classes to inherit sync functions and constants with minimal duplication in implementation and testers.

This refactor can generally be described as creating a `Base` counterpart (eg. `BaseClient` for `Client`) that takes all of the sync requirements, while `Client` inherits from `BaseClient` and implements functions that have reason to be async in an async implementation. The intention is to allow the addition of the async feature to have minimal impact on the existing implementation. This change allows us to inherit the bare minimum functionality for async classes while keeping testing streamlined and straightforward.

PR #70, opened 2020-06-25
refactor: create base collection class

rafilong:
towards #65

this PR is staged on top of #69 as refactor changes are cross-class; I encourage reviewing in sequential order


tests/unit/v1/test_collection.py
```diff
@@ -32,10 +32,18 @@ def _make_one(self, *args, **kwargs):
 
     @staticmethod
     def _get_public_methods(klass):
-        return set(
-            name
-            for name, value in six.iteritems(klass.__dict__)
-            if (not name.startswith("_") and isinstance(value, types.FunctionType))
+        return set().union(
```
crwilcox, line 35: why this change?

rafilong: The previous test was testing the functions that exist in `CollectionReference`. Now that we've pulled some of those out to `BaseCollectionReference`, we need to test the public interface, which means including the functions that exist in the parent. This was my way of solving that problem.

PR #88, opened 2020-07-09
test: add imports into test_collections systest

### LINKED ISSUE #84: Add System Test that gets collections
We were able to release v1.8.0 without a failure, despite missing an import. 

#83, #82 

We should add the following test 

```
db=firestore.Client()
collections = [i for i in db.collections()]
```

If that doesn't fail, 👍

IlyaFaer:
Closes #84

### DIFF
--- a/tests/system/test_system.py
+++ b/tests/system/test_system.py
@@ -68,12 +68,8 @@ def cleanup():
         operation()
 
 
-def test_collections():
-    from google.cloud import firestore
-
-    client = firestore.Client()
+def test_collections(client):
     collections = list(client.collections())
-
     assert isinstance(collections, list)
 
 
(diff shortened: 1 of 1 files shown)


tests/system/test_system.py
```diff
@@ -68,8 +68,12 @@ def cleanup():
         operation()
 
 
-def test_collections(client):
+def test_collections():
+    from google.cloud import firestore
+
+    client = firestore.Client()
     collections = list(client.collections())
+
```
IlyaFaer, line 76: We probably can just add an import into the existing test instead of writing the similar one

crwilcox: My concern here is it looks redundant and may get removed in the future. If I saw this I would wonder why I couldn't use the global firestore reference.

PR #104, opened 2020-07-16
refactor: drop six package use

IlyaFaer:
Closes #103

PR #106, opened 2020-07-16
refactor: drop six package use

### LINKED ISSUE #103: Remove use of six for v2
As we are removing python 2 support, I suspect we can remove use of six and remove a dependency from our library.

IlyaFaer:
Closes #103 
Supersedes #104

### DIFF
--- a/tests/unit/v1/test_base_query.py
+++ b/tests/unit/v1/test_base_query.py
@@ -16,14 +16,9 @@ import datetime
 import unittest
 
 import mock
-import six
 
 
 class TestBaseQuery(unittest.TestCase):
-
-    if six.PY2:
-        assertRaisesRegex = unittest.TestCase.assertRaisesRegexp
-
     @staticmethod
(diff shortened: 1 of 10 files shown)


crwilcox:
Thanks @IlyaFaer this generally lgtm. Adding Bu Sun as an FYI that we are trying to get rid of this if possible.

tests/unit/v1/test_base_query.py
```diff
@@ -252,7 +247,7 @@ def _where_unary_helper(self, value, op_enum, op_string="=="):
 
         field_pb = new_query._field_filters[0]
         expected_pb = StructuredQuery.UnaryFilter(
-            field=StructuredQuery.FieldReference(field_path=field_path), op=op_enum,
+            field=StructuredQuery.FieldReference(field_path=field_path), op=op_enum
```
rafilong, line 250: Not sure if this change is necessary, but at the same time I don't believe its problematic.

PR #113, opened 2020-07-17
fix: constructor invalid path tests

### LINKED ISSUE #102: Separate constructor invalid path test cases
Current [constructor tests](https://github.com/googleapis/python-firestore/blob/master/tests/unit/v1/test_collection.py#L64-L72) test multiple failure conditions in one test case. By separating these tests into multiple test cases it will be more clear which failure is the cause.

```
def test_constructor_invalid_path(self):
    with self.assertRaises(ValueError):
        self._make_one()
    with self.assertRaises(ValueError):
        self._make_one(99, "doc", "bad-collection-id")
    with self.assertRaises(ValueError):
        self._make_one("bad-document-ID", None, "sub-collection")
    with self.assertRaises(ValueError):
        self._make_one("Just", "A-Document")
```

rafilong:
Resolves #102


rafilong:
Continued in #114

PR #116, opened 2020-07-17
feat: use `DatetimeWithNanoseconds` throughout library

crwilcox:


tests/unit/v1/test_base_document.py
```diff
@@ -17,6 +17,8 @@
 import mock
 import datetime
```
rafilong, line 18: I think this is unused.

PR #118, opened 2020-07-20
feat: asyncio microgen client

### LINKED ISSUE #73: AsyncIO: Author AsyncIO Client

rafilong:
Resolves #73

### DIFF
--- a/google/cloud/firestore_v1/async_client.py
+++ b/google/cloud/firestore_v1/async_client.py
@@ -108,6 +108,7 @@ class AsyncClient(BaseClient):
     @property
     def _target(self):
         """Return the target (where the API is).
+        Eg. "firestore.googleapis.com"
 
         Returns:
             str: The location of the API.
--- a/google/cloud/firestore_v1/async_client.py
+++ b/google/cloud/firestore_v1/async_client.py
@@ -288,7 +288,7 @@ class AsyncClient(BaseClient):
             Sequence[:class:`~google.cloud.firestore_v1.async_collection.AsyncCollectionReference`]:
                 iterator of subcollections of the current document.
         """
-        iterator = self._firestore_api.list_collection_ids(
+        iterator = await self._firestore_api.list_collection_ids(
             request={"parent": "{}/documents".format(self._database_string)},
             metadata=self._rpc_metadata,
         )
@@ -297,7 +297,7 @@ class AsyncClient(BaseClient):
             for i in iterator.collection_ids:
                 yield self.collection(i)
             if iterator.next_page_token:
-                iterator = self._firestore_api.list_collection_ids(
+                iterator = await self._firestore_api.list_collection_ids(
                     request={
                         "parent": "{}/documents".format(self._database_string),
                         "page_token": iterator.next_page_token,
--- a/google/cloud/firestore_v1/async_client.py
+++ b/google/cloud/firestore_v1/async_client.py
@@ -99,33 +99,11 @@ class AsyncClient(BaseClient):
             :class:`~google.cloud.gapic.firestore.v1`.async_firestore_client.FirestoreAsyncClient:
             <The GAPIC client with the credentials of the current client.
         """
-        if self._firestore_api_internal is None:
-            # Use a custom channel.
-            # We need this in order to set appropriate keepalive options.
-
-            if self._emulator_host is not None:
-                # TODO(microgen): this likely needs to be adapted to use insecure_channel
-                # on new generated surface.
-                channel = firestore_grpc_transport.FirestoreGrpcAsyncIOTransport.create_channel(
-                    host=self._emulator_host
-                )
-            else:
-                channel = firestore_grpc_transport.FirestoreGrpcAsyncIOTransport.create_channel(
-                    self._target,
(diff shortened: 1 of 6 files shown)


tests/unit/v1/test_async_client.py
```diff
@@ -200,7 +201,8 @@ async def test_collections(self):
 
         collection_ids = ["users", "projects"]
         client = self._make_default_one()
-        firestore_api = mock.Mock(spec=["list_collection_ids"])
+        firestore_api = asyncmock.AsyncMock()
+        firestore_api.mock_add_spec(spec=["list_collection_ids"])
```
rafilong, line 205: If someone could take a look at this test case it would be appreciated. For some reason the final assertion in the test case is failing, stating that the mock is called 0 times, even though it is fairly evident the mock is being used from the fact that the previous assertions pass.

rafilong: Turns out there's some issue with `asyncmock` and using `assert_called_once_with`.

google/cloud/firestore_v1/base_client.py
```diff
@@ -117,11 +113,10 @@ def __init__(
         self._database = database
         self._emulator_host = os.getenv(_FIRESTORE_EMULATOR_HOST)
 
-    @property
-    def _firestore_api(self):
+    def _firestore_api_helper(self, transport, client_class, client_module):
         """Lazy-loading getter GAPIC Firestore API.
         Returns:
-            :class:`~google.cloud.gapic.firestore.v1`.firestore_client.FirestoreClient:
+            :class:`~google.cloud.gapic.firestore.v1`.async_firestore_client.FirestoreAsyncClient:
```
BenWhitehead, line 119: Isn't the type returned here dependant upon the subclass, either Async or not?

Is there a way we can list the or in the type annotation?

rafilong: That is a good point, and I'm not sure what the best way to resolve this is. Give me a bit to get back to you on that.

rafilong: From what it looks like [here](https://pythonhosted.org/an_example_pypi_project/sphinx.html#function-definitions), type is not required.

rafilong: I decided to remove type entirely, as `base_client` has no requirement on return type for the function. Its conceivable that at some point a client other than async and sync may exist, in which case the return type we specify would be outdated.

google/cloud/firestore_v1/async_client.py
```diff
@@ -86,6 +92,28 @@ def __init__(
...
+        """
+        return self._firestore_api_helper(
+            firestore_grpc_transport.FirestoreGrpcAsyncIOTransport,
+            firestore_client.FirestoreAsyncClient,
+            firestore_client,
+        )
+
+    @property
+    def _target(self):
+        """Return the target (where the API is).
```
tritone, line 110: What does target/location mean in this context? might be helpful to give an example.

tests/unit/v1/test__helpers.py
```diff
@@ -20,6 +20,11 @@
 import mock
 
 
+class AsyncMock(mock.MagicMock):
```
tritone, line 23: Why do you need to define your own AsyncMock rather than using https://docs.python.org/3/library/unittest.mock.html#unittest.mock.AsyncMock ?

google/cloud/firestore_v1/base_client.py
BenWhitehead: Fair, *grumbles about lack of scope constraints* :cry:

tests/unit/v1/test__helpers.py
rafilong: I didn't know that existed actually. Previously I was using AsyncMock provided by the [`asyncmock` package](https://pypi.org/project/asyncmock/), but it didn't work properly with some of the unittest features. I'll take a look at using this.

rafilong:
@tritone Unfortunately system tests do not exist yet - I was waiting to complete integration of the client before writing system tests, as I feel as though they are dependent on the entire client being converted to async, and it would be preferable to wait until that point. As this code is being committed to `v2-staging` rather than `master`, I feel like this is a reasonable choice.

google/cloud/firestore_v1/async_client.py
rafilong: Added example!

tests/unit/v1/test__helpers.py
rafilong: This library function is only available for Python 3.8 onwards, so unfortunately we can't use it. Is the current solution acceptable in this case?

tritone: Yep sounds good, thanks for looking into it.

tritone:
> @tritone Unfortunately system tests do not exist yet - I was waiting to complete integration of the client before writing system tests, as I feel as though they are dependent on the entire client being converted to async, and it would be preferable to wait until that point. As this code is being committed to `v2-staging` rather than `master`, I feel like this is a reasonable choice.

Discussed offline, I'm okay with merging this code into the branch before system tests are added.

PR #121, opened 2020-07-21
feat: asyncio microgen document

### LINKED ISSUE #77: AsyncIO: Author AsyncIO Document

rafilong:
Resolves #77

### DIFF
--- a/google/cloud/firestore_v1/async_document.py
+++ b/google/cloud/firestore_v1/async_document.py
@@ -284,7 +284,7 @@ class AsyncDocumentReference(BaseDocumentReference):
     async def get(self, field_paths=None, transaction=None):
         """Retrieve a snapshot of the current document.
 
-        See :meth:`~google.cloud.firestore_v1.client.Client.field_path` for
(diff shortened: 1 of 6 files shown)


google/cloud/firestore_v1/async_document.py
```diff
@@ -270,7 +270,7 @@ async def delete(self, option=None):
             still return the time that the request was received by the server.
         """
         write_pb = _helpers.pb_for_delete(self._document_path, option)
```
tritone, line 272: I wasn't on your earlier reviews but could you speak to how much of this code is duplicated from the existing document.py?

rafilong: There's a fair amount of duplication. As far as shared functionality goes, most was refactored to the base class, but instances where rpc calls are made with the same parameters but with an async endpoint, there isn't much we can do to avoid duplication of parameters.

rafilong: If construction of parameters is in any way complex, we can refactor construction to a helper function in the base class.

rafilong:
Nice catch, for the most part the docstrings were good, but there was one function that I overlooked during my refactor.

PR #123, opened 2020-07-22
feat: asyncio microgen transaction

### LINKED ISSUE #78: AsyncIO: Author AsyncIO Transaction

rafilong:
Resolves #78

### DIFF
--- a/google/cloud/firestore_v1/async_transaction.py
+++ b/google/cloud/firestore_v1/async_transaction.py
@@ -85,7 +85,7 @@ class AsyncTransaction(async_batch.AsyncWriteBatch, BaseTransaction):
             msg = _CANT_BEGIN.format(self._id)
             raise ValueError(msg)
 
-        transaction_response = self._client._firestore_api.begin_transaction(
+        transaction_response = await self._client._firestore_api.begin_transaction(
             request={
                 "database": self._client._database_string,
                 "options": self._options_protobuf(retry_id),
@@ -105,7 +105,7 @@ class AsyncTransaction(async_batch.AsyncWriteBatch, BaseTransaction):
 
         try:
             # NOTE: The response is just ``google.protobuf.Empty``.
-            self._client._firestore_api.rollback(
+            await self._client._firestore_api.rollback(
                 request={
                     "database": self._client._database_string,
                     "transaction": self._id,
@@ -148,7 +148,7 @@ class AsyncTransaction(async_batch.AsyncWriteBatch, BaseTransaction):
             .DocumentSnapshot: The next document snapshot that fulfills the
             query, or :data:`None` if the document does not exist.
         """
-        return self._client.get_all(references, transaction=self)
+        return await self._client.get_all(references, transaction=self)
 
     async def get(self, ref_or_query):
         """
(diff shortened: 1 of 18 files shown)


tests/unit/v1/test_async_transaction.py
```diff
@@ -80,15 +82,10 @@ def test__clean_up(self):
...
     @pytest.mark.asyncio
     async def test__begin(self):
-        from google.cloud.firestore_v1.services.firestore import (
-            client as firestore_client,
-        )
         from google.cloud.firestore_v1.types import firestore
 
         # Create a minimal fake GAPIC with a dummy result.
-        firestore_api = mock.create_autospec(
-            firestore_client.FirestoreClient, instance=True
```
tritone, line 90: So my understanding is that using `create_autospec` implies that the mock's methods will exactly match those of the FirestoreClient, whereas using a generic AsyncMock allows you to call any method, including ones that don't exist on the real client. Any thoughts on whether this will be a reasonable choice for this test (and the ones below with similar changes)?

rafilong: That is a valid point - my thought was that as functionality is only defined for the mocked functions we care about, that autospecing is not entirely necessary. I can see what you're coming from though, and am happy to manually add spec. @crwilcox what are your thoughts?

crwilcox: @tritone has it right here. Autospecs will fail if you try to call a method that doesn't match the class it is mocking. On the other hand, it often means more setup, when a generic mock would allow it in less.

I think autospecs can be useful. Is there a reason to move to a generic mock that I am not seeing?

rafilong: We're using the modified mock class (`AsyncMock`) that overloads the `__call__` function with a coroutine. This isn't stopping us from manually creating the spec, but I don't believe we can use autospec.

google/cloud/firestore_v1/async_transaction.py
```diff
@@ -148,7 +148,7 @@ async def get_all(self, references):
             .DocumentSnapshot: The next document snapshot that fulfills the
             query, or :data:`None` if the document does not exist.
         """
-        return self._client.get_all(references, transaction=self)
+        return await self._client.get_all(references, transaction=self)
```
crwilcox, line 151: It appears that this pr would change the return type from a coroutine to the result. Was this a bug?

tests/unit/v1/test_async_transaction.py
crwilcox: We talked about this a bit. We already use a mix so this matches precedent. I do think autospec may have value but also isn't necessary. I would be satisfied to track this as an issue to dig into later.

google/cloud/firestore_v1/async_transaction.py
rafilong: Yep, I hadn't realized that the rpc call in the microgen client returned a coroutine.

tests/unit/v1/test_async_transaction.py
rafilong: Created issue #133

tritone: Sounds good to me; thanks for following up.

Issue #131, opened 2020-07-28
Add Watch to async interface


rafilong:
Initially removed `Watch` from async interface due to lack if async support with `ResumableBidiRPC`. Preferably we can add support back before v2 release.

rafilong:
https://github.com/googleapis/python-firestore/pull/132#discussion_r461873402 note on adding `_transport` interface to client.

Prince25:
Just curious about the progress on this. It's been more than a year. It's currently halting progress on a WebRTC application so I must use an alternative to Firebase.

JJ11teen:
Hey team, any progress or possible roadmap to share with this one? (:

NinitoAS:
Hi,

Any updates on this? It's almost been 4.5 years since someone requested this. Is Q2 2024 a possible milestone?

This is a game-changer for us and I hope it's a priority for the team

NinitoAS:
Following up again as we have not heard anything else. Any updates?

n00bsys0p:
This would also be most useful for our team. Is it even feasible to implement, or are there major blockers that make it unlikely?

NinitoAS:
Any updates?

PR #132, opened 2020-07-28
feat: asyncio system tests

rafilong:
Added system tests and removed `Watch` from interface. Opened issue to re-add `Watch` support at a later date #131.

Additionally, made fixes to improperly mocked return types in `AsyncClient.list_documents` and `AsyncQuery.stream`.

This PR is stacked on #123.


google/cloud/firestore_v1/async_collection.py
```diff
@@ -119,7 +117,8 @@ async def list_documents(self, page_size=None):
             },
             metadata=self._client._rpc_metadata,
         )
-        return (_item_to_document_ref(self, i) for i in iterator)
+        async for i in iterator:
+            yield _item_to_document_ref(self, i)
```
rafilong, line 121: Prior to this change `list_documents` was used like `[i async for i in await list_documents]`. This change simplifies the use case to `[i async for i in list_documents]`. Previous type was a coroutine that returned an async generator, current type is just an async generator.

Alternatively, we could compose the entire list and return just that, though that seems counterproductive.

google/cloud/firestore_v1/__init__.py
```diff
@@ -100,6 +109,13 @@
     "__version__",
     "ArrayRemove",
     "ArrayUnion",
+    "AsyncClient",
+    "AsyncCollectionReference",
+    "AsyncDocumentReference",
+    "AsyncQuery",
+    "AsyncTransaction",
+    "async_transactional",
```
crwilcox, line 117: nit: I think these may be alpha sorted?

google/cloud/firestore_v1/async_client.py
```diff
@@ -99,11 +99,15 @@ def _firestore_api(self):
...
         """
-        return self._firestore_api_helper(
+        firestore_api = self._firestore_api_helper(
             firestore_grpc_transport.FirestoreGrpcAsyncIOTransport,
             firestore_client.FirestoreAsyncClient,
             firestore_client,
         )
+        # Add _transport to FirestoreAsyncClient interface for use in Watch
+        if hasattr(firestore_api, "_client"):
+            firestore_api._transport = firestore_api._client._transport
```
crwilcox, line 109: Rather than modifying the generated object instance, could we do this same logic, but in watch? If in multiple spots, maybe make an internal lookup? Or add this as a property on the client (our manual one)?

google/cloud/firestore_v1/async_collection.py
crwilcox: I would suggest we use a generator. ( `( )` instead of `[ ]` )

tests/unit/v1/test_async_collection.py
```diff
@@ -206,17 +206,19 @@ def _next_page(self):
...
         collection = self._make_one("collection", client=client)
 
         if page_size is not None:
-            documents = list(await collection.list_documents(page_size=page_size))
+            documents = [
+                i async for i in collection.list_documents(page_size=page_size)
+            ]
         else:
-            documents = list(await collection.list_documents())
+            documents = [i async for i in collection.list_documents()]
```
crwilcox, line 221: Is this change required? It seems at first glance it should be equivalent?

tests/system/test_system_async.py
```diff
@@ -0,0 +1,1000 @@
...
+from google.api_core.exceptions import AlreadyExists
+from google.api_core.exceptions import FailedPrecondition
+from google.api_core.exceptions import InvalidArgument
+from google.api_core.exceptions import NotFound
+from google.cloud._helpers import _datetime_to_pb_timestamp
+from google.cloud._helpers import UTC
+from google.cloud import firestore_v1 as firestore
+from test_utils.system import unique_resource_id
+
+FIRESTORE_CREDS = os.environ.get("FIRESTORE_APPLICATION_CREDENTIALS")
```
BenWhitehead, line 34: Most projects use the environment variable `GOOGLE_APPLICATION_CREDENTIALS` to locate the service account to use.

@crwilcox Is this accurate for python projects?

tests/system/test_system_async.py
```diff
@@ -0,0 +1,1000 @@
...
+    yield firestore.AsyncClient(project=project, credentials=credentials)
+
+
+@pytest.fixture
+async def cleanup():
+    operations = []
+    yield operations.append
+
+    for operation in operations:
+        await operation()
```
BenWhitehead, line 58: do each of these `await`s need to be guarded for possible exceptions? We probably don't want to let one `operation` failing preventing the rest from being attempted.

tests/unit/v1/test_async_collection.py
rafilong: You would think so, but `list()` can't be used to create a list from an async generator, as async generators are not iterable. Unfortunately, this is the preferred solution to creating a list from an async generator.

google/cloud/firestore_v1/__init__.py
rafilong: Good point, I thought it might make sense to group by function but alpha sorting is probably more standard.

google/cloud/firestore_v1/async_client.py
rafilong: If we create an `AsyncWatch`, this definitely something that can be moved there. In fact, I should have most likely removed this when I removed `Watch` from the async interface. I'm not quite sure what you mean by your last point - isn't this as close as we can get to adding this as a property to the manual client?

tests/unit/v1/test_async_collection.py
rafilong: If we want to use something like `list(await collection.list_documents())` we could refactor `list_documents` to return a list, rather than an async generator. However, we would lose on the benefit of a generator being able to incrementally return results as they arrive, rather than compiling all results before returning.

google/cloud/firestore_v1/async_collection.py
rafilong: Resolved in chat. Using a generator results in a different return type, which complicates the use case. In the case that we do not explicitly use yield, we have a coroutine that returns an async generator, which must be awaited on before using `async for`. If we explicitly iterate over the results and yield, the type is an async generator which we can directly use `async for` on.

tests/system/test_system_async.py
```diff
@@ -0,0 +1,1000 @@
...
+from google.cloud._helpers import UTC
+from google.cloud import firestore_v1 as firestore
+from test_utils.system import unique_resource_id
+
+FIRESTORE_CREDS = os.environ.get("FIRESTORE_APPLICATION_CREDENTIALS")
+FIRESTORE_PROJECT = os.environ.get("GCLOUD_PROJECT")
+RANDOM_ID_REGEX = re.compile("^[a-zA-Z0-9]{20}$")
+MISSING_DOCUMENT = "No document to update: "
+DOCUMENT_EXISTS = "Document already exists: "
+UNIQUE_RESOURCE_ID = unique_resource_id("-")
```
BenWhitehead, line 39: Is this id unique to the scope of all tests, or unique for each test (I'm not familiar with the lifecycle of pytest files)?

I think we'll want this to be unique for each test, correct? If it's not already scoped that way we might need to move this to the fixture so that it is.

google/cloud/firestore_v1/async_client.py
crwilcox: I don't like modifying the gen object if possible. Either by adding a property (that does this) to the manual client or moving to watch would make sense to me though.

tests/system/test_system_async.py
crwilcox: @BenWhitehead  yep, specifically for firestore. this way we can use the same kokoro config :) The project that is `GOOGLE_APPLICATION_CREDENTIALS` is datastore

tests/system/test_system_async.py
rafilong: This fixture is used to delete documents and collections after the test is run, so I believe failing here is acceptable. The awaits only happen after the test is run, and failure should not directly impact other tests.

tests/system/test_system_async.py
rafilong: From what I remember, Firestore uses a different environment variable than other Python projects. This is consistent with the sync version of Firestore.

tests/system/test_system_async.py
rafilong: This is something I mimicked from the sync system tests. I believe it is unique to the scope of all tests, though each test individually defines a prefix to make the id unique for each test. Is there a compelling reason to generated a unique id postfix for all tests?

google/cloud/firestore_v1/async_client.py
rafilong: Sounds good, will keep in mind for the `Watch` PR. Added comment to open issue.

tests/system/test_system_async.py
crwilcox: I think borrowing from sync should be fine here.

tests/system/test_system_async.py
BenWhitehead: sounds good

tests/system/test_system_async.py
BenWhitehead: Sounds good

PR #145, opened 2020-08-06
feat: integrate limit to last

rafilong:
Integrate changes from #57 to the async client.

### DIFF
--- a/google/cloud/firestore_v1/async_collection.py
+++ b/google/cloud/firestore_v1/async_collection.py
@@ -13,9 +13,6 @@
 # limitations under the License.
 
 """Classes for representing collections for the Google Cloud Firestore API."""
-import warnings
-
-
 from google.cloud.firestore_v1.base_collection import (
     BaseCollectionReference,
     _auto_id,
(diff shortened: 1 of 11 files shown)


google/cloud/firestore_v1/async_collection.py
```diff
@@ -130,17 +127,26 @@ async def list_documents(
...
-            "'Collection.get' is deprecated:  please use 'Collection.stream' instead.",
-            DeprecationWarning,
-            stacklevel=2,
-        )
-        async for d in self.stream(transaction=transaction):
-            yield d  # pytype: disable=name-error
+    async def get(self, transaction=None) -> list:
+        """Read the documents in this collection.
+
+        This sends a ``RunQuery`` RPC and returns a list of documents
```
crwilcox, line 133: So, this deprecated bit makes me wonder if we shouldn't include this. Is that perhaps why it isn't here?

google/cloud/firestore_v1/async_query.py
```diff
@@ -105,23 +105,47 @@ def __init__(
...
-            DeprecationWarning,
-            stacklevel=2,
-        )
-        async for d in self.stream(transaction=transaction):
-            yield d
+    async def get(self, transaction=None) -> list:
+        """Read the documents in the collection that match this query.
+        This sends a ``RunQuery`` RPC and returns a list of documents
+        returned in the stream of ``RunQueryResponse`` messages.
+        Args:
```
crwilcox, line 119: same comment.

google/cloud/firestore_v1/collection.py
```diff
@@ -121,14 +119,22 @@ def list_documents(self, page_size=None) -> Generator[Any, Any, None]:
...
-    def get(self, transaction=None) -> Generator[document.DocumentSnapshot, Any, None]:
-        """Deprecated alias for :meth:`stream`."""
-        warnings.warn(
-            "'Collection.get' is deprecated:  please use 'Collection.stream' instead.",
-            DeprecationWarning,
-            stacklevel=2,
-        )
-        return self.stream(transaction=transaction)
+    def get(self, transaction=None) -> list:
+        """Read the documents in this collection.
```
crwilcox, line 123: oh, wait, did we undeprecate this? memory is fuzzy.

tests/unit/v1/test_async_collection.py
```diff
@@ -249,47 +249,27 @@ async def test_list_documents_w_page_size(self):
...
-
-        query_class.return_value.stream.return_value = AsyncIter(range(3))
-
         collection = self._make_one("collection")
-        with warnings.catch_warnings(record=True) as warned:
-            get_response = collection.get()
-
-            async for _ in get_response:
-                pass
+        get_response = await collection.get()
```
crwilcox, line 253: how does this map to the sync version?

google/cloud/firestore_v1/collection.py
rafilong: Yep, it looks like it was undeprecated after the `v2-staging` split.

tests/unit/v1/test_async_collection.py
rafilong: Same functionality, now neither are Iterator/Generator/AsyncGenerators. Both aggregate results so they can be reversed and return a list, though the async function is a coroutine that has that functionality.

google/cloud/firestore_v1/collection.py
rafilong: As seen on [master](https://github.com/googleapis/python-firestore/blob/master/google/cloud/firestore_v1/query.py#L784-L820)

google/cloud/firestore_v1/async_query.py
rafilong: See following comment.

google/cloud/firestore_v1/async_collection.py
rafilong: See following comment.

PR #176, opened 2020-08-27
feat: improve type information

### LINKED ISSUE #139: Improve type information from `Any`
https://github.com/googleapis/python-firestore/pull/134 adds type hints to the manual surface.  Though in many cases, the type specified is `Any` which isn't all that helpful. It would be great if we could remove instances of `Any` and replace with a concrete type.

if you search for `-> Any` all of those could possibly improved with more precise type hints.

Work should target `v2-staging` branch for the time being.

HemangChothani:
Fixes #139


tseaver:
@HemangChothani Please rebase / retarget your changes against the current `master` branch.  The `v2-staging` branch is no longer active (@crwilcox  force-pushed it to `master` last week).

tseaver:
Blocked on #188.

google/cloud/firestore_v1/_helpers.py
```diff
@@ -921,7 +926,7 @@ class ReadAfterWriteError(Exception):
     """
 
 
-def get_transaction_id(transaction, read_operation=True) -> Any:
+def get_transaction_id(transaction, read_operation=True) -> Optional[bytes]:
```
tseaver, line 929: I didn't think `Optional` was appropriate for return values, but according to [the docs](https://docs.python.org/3/library/typing.html#typing.Optional), it seems like it might be so.  It would, however, seem clearer as:
```suggestion
def get_transaction_id(transaction, read_operation=True) -> Union[bytes, None]:
```

google/cloud/firestore_v1/base_document.py
```diff
@@ -355,7 +358,7 @@ def get(self, field_path) -> Any:
         nested_data = field_path_module.get_nested_value(field_path, self._data)
         return copy.deepcopy(nested_data)
 
-    def to_dict(self) -> Any:
+    def to_dict(self) -> Optional[Dict[str, Any]]:
```
tseaver, line 361: Likewise, this would be clearer as:
```suggestion
    def to_dict(self) -> Union[Dict[str, Any], None]:
```

google/cloud/firestore_v1/document.py
```diff
@@ -76,7 +76,7 @@ def create(self, document_data) -> Any:
         write_results = batch.commit()
         return _first_write_result(write_results)
 
-    def set(self, document_data, merge=False) -> Any:
+    def set(self, document_data, merge=False) -> Coroutine:
```
tseaver, line 79: This method returns a single `google.cloud.proto.firestore.v1.write.WriteResult` instance.
```suggestion
    def set(self, document_data, merge=False) -> write.WriteResult:
```

google/cloud/firestore_v1/document.py
```diff
@@ -107,7 +107,7 @@ def set(self, document_data, merge=False) -> Any:
         write_results = batch.commit()
         return _first_write_result(write_results)
 
-    def update(self, field_updates, option=None) -> Any:
+    def update(self, field_updates, option=None) -> Coroutine:
```
tseaver, line 110: This method returns a single `google.cloud.proto.firestore.v1.write.WriteResult` instance.
```suggestion
    def update(self, field_updates, option=None) -> write.WriteResult:
```

google/cloud/firestore_v1/document.py
```diff
@@ -255,7 +255,7 @@ def update(self, field_updates, option=None) -> Any:
         write_results = batch.commit()
         return _first_write_result(write_results)
 
-    def delete(self, option=None) -> Any:
+    def delete(self, option=None) -> Coroutine:
```
tseaver, line 258: This method returns a single `google.protobuf.timestamp_pb2.Timestamp` instance.
```suggestion
    def set(self, option=None) -> timestamp_pb2.WriteResult:
```

google/cloud/firestore_v1/transaction.py
```diff
@@ -136,7 +140,7 @@ def _commit(self) -> list:
         self._clean_up()
         return list(commit_response.write_results)
 
-    def get_all(self, references) -> Any:
+    def get_all(self, references) -> Coroutine:
```
tseaver, line 143: Should be a generator:
```suggestion
    def get_all(self, references) -> Generator[DocumentSnapshot, Any, Any]:  # send type None?
```

google/cloud/firestore_v1/transaction.py
```diff
@@ -182,7 +186,7 @@ class _Transactional(_BaseTransactional):
     def __init__(self, to_wrap) -> None:
         super(_Transactional, self).__init__(to_wrap)
 
-    def _pre_commit(self, transaction, *args, **kwargs) -> Any:
+    def _pre_commit(self, transaction, *args, **kwargs) -> Coroutine:
```
tseaver, line 189: Here `Any` is correct:  it will be whatever type the wrapped callable returns.

tseaver (commented):
@crwilcox Please check my comments here:  my typing-fu may not be strong enough.

google/cloud/firestore_v1/document.py
HemangChothani: I have already tried that but `pytype` throws an error like `Expected: google.cloud.firestore_v1.types.write.WriteResult
  Actually returned: coroutine`
Also found used `coroutine` in async_document.py file
https://github.com/googleapis/python-firestore/blob/2021f38bb6f016c13bc43d59730c77b57ae5c352/google/cloud/firestore_v1/async_document.py#L78

google/cloud/firestore_v1/document.py
HemangChothani: Same error as mentioned in above comment

crwilcox:
@tseaver I think your comments are generally correct. Not sure my stance on optional vs union with none, but they should be ~equivalent.

HemangChothani:
@tseaver  PTAL

crwilcox:
@tseaver I think your suggestions make sense. Unfortunately they are out of date so I cannot auto-apply

google/cloud/firestore_v1/base_document.py
```diff
@@ -188,12 +188,10 @@ def create(self, document_data: dict) -> NoReturn:
     def set(self, document_data: dict, merge: bool = False) -> NoReturn:
         raise NotImplementedError
 
-    def update(
-        self, field_updates: dict, option: _helpers.WriteOption = None
-    ) -> NoReturn:
+    def update(self, field_updates: dict, option=None) -> NoReturn:
```
tseaver, line 191: Why are we going back on the annotations here?
```suggestion
    def update(
        self, field_updates: dict, option: _helpers.WriteOption = None
    ) -> NoReturn:
```

google/cloud/firestore_v1/base_document.py
```diff
@@ -188,12 +188,10 @@ def create(self, document_data: dict) -> NoReturn:
...
         raise NotImplementedError
 
-    def delete(self, option: _helpers.WriteOption = None) -> NoReturn:
+    def delete(self, option=None) -> NoReturn:
```
tseaver, line 194: Likewise here?

tseaver:
```suggestion
   def delete(self, option: _helpers.WriteOption = None) -> NoReturn:
```

google/cloud/firestore_v1/document.py
```diff
@@ -26,7 +26,7 @@
 from google.cloud.firestore_v1.types import write
 from google.cloud.firestore_v1.watch import Watch
 from google.protobuf import timestamp_pb2
-from typing import Any, Generator
+from typing import Any, Callable, Generator, Iterable
```
tseaver, line 29: I guess `Callable` and `Iterable` are being used in code from the other commit?

google/cloud/firestore_v1/transaction.py
```diff
@@ -36,11 +36,11 @@
 from google.cloud.firestore_v1 import batch
 from google.cloud.firestore_v1.document import DocumentReference
 from google.cloud.firestore_v1.query import Query
-from typing import Any, Generator, Optional
 
 # Types needed only for Type Hints
 from google.cloud.firestore_v1.base_document import DocumentSnapshot
 from google.cloud.firestore_v1.types import CommitResponse
+from typing import Any, Callable, Generator, Optional
```
tseaver, line 43: Likewise here?

tseaver (changes requested):

google/cloud/firestore_v1/document.py
HemangChothani: Yes.

google/cloud/firestore_v1/transaction.py
HemangChothani: Yes

google/cloud/firestore_v1/base_document.py
HemangChothani: When run the unit test it throws an error like `AttributeError: module 'google.cloud.firestore_v1._helpers' has no attribute 'WriteOption'`

HemangChothani: Solved the error after removing import of `DocumentReference` from _helper.py file

google/cloud/firestore_v1/_helpers.py
```diff
@@ -222,7 +219,7 @@ def encode_dict(values_dict) -> dict:
     return {key: encode_value(value) for key, value in values_dict.items()}
 
 
-def reference_value_to_document(reference_value, client) -> DocumentReference:
+def reference_value_to_document(reference_value, client) -> Any:
```
tseaver: Ugh, we have cyclic imports here.  Could we use a string to get a better type hint?  E.g.:
```python
def reference_value_to_document(reference_value, client) -> "google.cloud.firestore_v1.document.DocumentReference":
```

HemangChothani: It still throws an error, for string also need to import the class.

tseaver (approved):

crwilcox (approved):

Issue #190, opened 2020-09-18
Restore unit test coverage to 100%


tseaver:
See: https://github.com/googleapis/python-firestore/pull/187#issuecomment-695036078

tseaver:
Duplicate of #92.

PR #191, opened 2020-09-18
tests: drop min coverage to 97%

tseaver:
Toward https://github.com/googleapis/python-firestore/issues/190

Note that this is a short-term fix, to get builds against `master`
passing while work goes on to bring coverage back to 100%.


google/cloud/firestore_admin_v1/services/firestore_admin/async_client.py
```diff
@@ -273,9 +273,9 @@ async def list_indexes(
                 maximum=60.0,
                 multiplier=1.3,
                 predicate=retries.if_exception_type(
-                    exceptions.DeadlineExceeded,
```
busunkim96, line 276: Is the generator going back and forth on the order of these exceptions?  https://github.com/googleapis/gapic-generator-python/blob/b19026d9cca26ebd1cd0c3e73f738c4d1870d987/gapic/templates/%25namespace/%25name_%25version/%25sub/services/%25service/async_client.py.j2#L201-L206

synth.py
```diff
@@ -179,6 +179,33 @@ def lint_setup_py(session):
...
+# Work around https://github.com/googleapis/python-firestore/issues/190
+
+s.replace(
+    "noxfile.py",
+    """\
+    session.run\("coverage", "report", "--show-missing", "--fail-under=100"\)""",
+    """\
+    # Work around https://github.com/googleapis/python-firestore/issues/190
+    session.run("coverage", "report", "--show-missing", "--fail-under=97")""",
+)
```
busunkim96, line 191: You can pass coverage to the `py_library` and it will adjust the line in noxfile.py

```py
templated_files = common.py_library(
    samples=False,  # set to True only if there are samples
    microgenerator=True,
    cov_level=9,7
)
```

tseaver: 8fda6aa

busunkim96 (approved):

google/cloud/firestore_admin_v1/services/firestore_admin/async_client.py
tseaver: I don't understand how that template fragment could be varying the output.

Issue #95, opened 2020-07-14
Run Cross Language tests for v2


crwilcox:
tests/unit/v1/test_cross_language.py → tests/unit/v1/_test_cross_language.py 

remove the leading `_` and begin testing with these again. Disabled as the protos for this need to be generated with protoplus to run correctly. (see makefile)

HemangChothani:
@crwilcox  I have generated `test.proto` file as per mentioned in `makefile`, so should i open a PR `https://github.com/googleapis/googleapis`  here with new generated test.proto file?

crwilcox:
@HemangChothani you modified the makefile to generate using protoplus?

HemangChothani:
@crwilcox  No, i didn't make any changes in makefile, i had follow the steps and generate `test_pb2.py` file, i don't know what to make changes in makefile related to `protoplus`, i have tried to find any related documentation, but not succeed.

PR #205, opened 2020-09-25
tests: re-enable cross-language conformance tests

tseaver:
~~WIP:  still need to actually *fix* the tests, but they are at least running.~~

Closes: #95


Makefile_v1
```diff
@@ -11,30 +11,47 @@ GOOGLEAPIS_REPO = $(REPO_DIR)/googleapis
 TESTS_REPO = $(REPO_DIR)/conformance-tests
 TEST_PROTO_DIR = $(TESTS_REPO)/firestore/v1
 TEST_PROTO_SRC = $(TEST_PROTO_DIR)/proto/google/cloud/conformance/firestore/v1/tests.proto
+TESTDATA_DIR = `pwd`/tests/unit/v1/testdata/
 
 TMPDIR = /tmp/python-fs-proto
-TMPDIR_FS = $(TMPDIR)/google/cloud/firestore_v1/proto
+TMPDIR_FS = $(TMPDIR)/google/cloud/firestore_v1/types
 TEST_PROTO_COPY = $(TMPDIR_FS)/tests.proto
```
tseaver, line 18: Hmm, I'm wondering if the `tests.proto`-derived module should live somewhere else, e.g., maybe in `tests/unit/V1`?  It's name is unfortunate, if we do that (maybe rename it to `conftests.proto`?

tests/unit/v1/test_cross_language.py
```diff
@@ -125,11 +106,17 @@ def _run_testcase(testcase, call, firestore_api, client):
             call()
     else:
         call()
+
+        wrapped_writes = [write.Write(write_pb) for write_pb in testcase.request.writes]
```
tseaver, line 110: Hmm, maybe this should use `Message.wrap`?  E.g.:
```suggestion
        wrapped_writes = [write.Write.wrap(write_pb) for write_pb in testcase.request.writes]
```

tests/unit/v1/test_cross_language.py
```diff
@@ -245,17 +238,20 @@ def callback(keys, applied_changes, read_time):
                 db_str = "projects/projectID/databases/(default)"
                 watch._firestore._database_string_internal = db_str
 
+                wrapped_responses = [
+                    firestore.ListenResponse(proto) for proto in testcase.responses
```
tseaver, line 242: Likewise here: 
```suggestion
                    firestore.ListenResponse.wrap(proto) for proto in testcase.responses
```

Makefile_v1
crwilcox: I would be alright with either change, if only because we don't need to ship protos otherwise.

tests/unit/v1/test_cross_language.py
crwilcox: is this the cause of bad __eq__?

tseaver: Yes.  The conftest protos have the expected requests laid out as "raw" proto-generated classes, which don't compare equal to our wrapped classes (asymmetric).

Makefile_v1
tseaver: I think I'll go ahead and move it.

tseaver: Hmm, the Docker version of the generator (`gcr.io/gapic-images/gapic-generator-python`) doesn't seem to find the `tests.proto` file if I move it out of `google/`.

tseaver:
@crwilcox PTAL.

crwilcox (approved):

PR #209, opened 2020-09-25
WIP: First stab at partition queries

chrisrossi:
Rough draft, synchronous API only.

### DIFF
--- a/google/cloud/firestore_v1/query.py
+++ b/google/cloud/firestore_v1/query.py
@@ -27,8 +27,11 @@ from google.cloud.firestore_v1.base_query import (
 
 from google.cloud.firestore_v1 import _helpers
 from google.cloud.firestore_v1 import document
+from google.cloud.firestore_v1 import field_path
 from google.cloud.firestore_v1.watch import Watch
 from typing import Any, Generator
+
+DOCUMENT_ID = field_path.FieldPath.document_id()
 
 
 class Query(BaseQuery):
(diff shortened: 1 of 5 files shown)


google/cloud/firestore_v1/query.py
```diff
@@ -239,3 +242,96 @@ def on_snapshot(docs, changes, read_time):
...
+
+    def _partition(self, start_at, end_before):
+        """TODO"""
+        if start_at:
+            start_at = start_at, True
+
+        if end_before:
+            end_before = end_before, True
+
+        return Query(
```
chrisrossi, line 327: The Java implementation returns a `QueryPartition` object with a `create_query` method. On first blush, it seemed like that intermediate step was probably unnecessary, so I'm marshaling query objects here. Although, it occurs to me that if we're using this in some kind of distributed map/reduce paradigm, that an intermediate form that provides a serializable cursor, might make sense.

google/cloud/firestore_v1/query.py
```diff
@@ -239,3 +242,96 @@ def on_snapshot(docs, changes, read_time):
...
+            self._parent,
+            orders=(Query._make_order(DOCUMENT_ID, Query.ASCENDING),),
+            all_descendants=True,
+            start_at=start_at,
+            end_at=end_before,
+        )
+
+    def _decode_cursor(self, cursor_pb):
+        """TODO"""
+        return self._client.document(cursor_pb.values[0].reference_value).get()
```
chrisrossi, line 337: I don't like that we're having to retrieve the document just to get the cursor into the format that `BaseQuery` is expecting.  I'm planning to look into ways to get around this. It should be as simple as handing the cursor back to Firestore in more or less the same way it came to us.

tests/system/test_system.py
```diff
@@ -52,7 +53,7 @@ def _get_credentials_and_project():
     return credentials, project
 
 
-@pytest.fixture(scope=u"module")
+@pytest.fixture(scope="module")
```
chrisrossi, line 56: Blacken must have done this.

tests/system/test_system.py
```diff
@@ -890,6 +891,58 @@ def test_collection_group_queries_filters(client, cleanup):
...
+        cleanup_batch.delete(doc_ref)
+
+    cleanup(cleanup_batch.commit)
+    batch.commit()
+
+    query = client.collection_group(collection_group)
+    streams = [partition.stream() for partition in query.get_partitions(3)]
+    snapshots = itertools.chain(*streams)
+    found = [snapshot.id for snapshot in snapshots]
+    found.sort()
```
chrisrossi, line 941: Order that documents are returned in is odd and seems to be related to underlying implementation details.

For example, if docs are number sequentially starting at 0 and we get back 3 cursors (4 partitions), the first partition will be:

1, 5, 9, 13, ..., 381, 0, 4, 8, 12

I assume that's expected and we shouldn't monkey with it.

google/cloud/firestore_v1/query.py
```diff
@@ -239,3 +242,96 @@ def on_snapshot(docs, changes, read_time):
...
+            limit_to_last=limit_to_last,
+            offset=offset,
+            start_at=start_at,
+            end_at=end_at,
+            all_descendants=all_descendants,
+        )
+
+    def get_partitions(self, partition_count) -> Generator[Query, None, None]:
+        """TODO"""
+        # Validate query?
```
chrisrossi, line 287: I'm pretty sure the queries this will work on is pretty restricted. I'm not sure what all the restrictions are at this point. I can and probably will write some tests to probe what the limitations are, but they might also be written down somewhere.

tests/system/test_system.py
tseaver: 'SOK, especially given that firestore is now PY3-only.

tseaver: Wait, was that from running `nox -e blacken`?  Because that doesn't remove the `u` prefixes on my machine.

chrisrossi: Yes. Here's what I happens when I reintroduce one and then run `nox -e blacken`:

~~~~
(firestore) chris@spirit:~/proj/google/firestore$ git diff
diff --git a/tests/system/test_system.py b/tests/system/test_system.py
index e68e31c..68a2a7b 100644
--- a/tests/system/test_system.py
+++ b/tests/system/test_system.py
@@ -53,7 +53,7 @@ def _get_credentials_and_project():
     return credentials, project
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture(scope=u"module")
 def client():
     credentials, project = _get_credentials_and_project()
     yield firestore.Client(project=project, credentials=credentials)
(firestore) chris@spirit:~/proj/google/firestore$ nox -e blacken
nox > Running session blacken
nox > Creating virtual environment (virtualenv) using python3.6 in .nox/blacken
nox > pip install black==19.10b0
nox > black docs google tests noxfile.py setup.py
reformatted /home/chris/proj/google/firestore/tests/system/test_system.py
All done! ✨ 🍰 ✨
1 file reformatted, 97 files left unchanged.
nox > Session blacken was successful.
(firestore) chris@spirit:~/proj/google/firestore$ git diff
(firestore) chris@spirit:~/proj/google/firestore$ 
~~~~

chrisrossi:
Had a misfire, sorry. Will reopen later today.

PR #210, opened 2020-10-02
feat: partition queries

chrisrossi:
Implement the new partition queries feature for Firestore.


google/cloud/firestore_v1/query.py
```diff
@@ -239,3 +241,93 @@ def on_snapshot(docs, changes, read_time):
...
+        field_filters=(),
+        orders=(),
+        limit=None,
+        limit_to_last=False,
+        offset=None,
+        start_at=None,
+        end_at=None,
+        all_descendants=True,
+    ) -> None:
+        # TODO: Validate all_descendants?
```
cguardia, line 271: Maybe not? Doesn't look like BaseQuery does that either.

cguardia (approved):
Looks good.

PR #241, opened 2020-11-03
chore: add trove classifier to setup.py for 3.8, 3.9

crwilcox:
release-as: 2.0.0


.github/workflows/prerelease.yml
```diff
@@ -0,0 +1,23 @@
+name: Python-pre-releases
+
+on:
+  pull_request:
+  schedule:
```
crwilcox, line 5: FYI, likely doesn't need to run on each PR, but if this isn't very expensive, it may make sense as our other tests take longer.

BenWhitehead (approved):

noxfile.py
```diff
@@ -27,10 +27,9 @@
 BLACK_VERSION = "black==19.10b0"
 BLACK_PATHS = ["docs", "google", "tests", "noxfile.py", "setup.py"]
 
-DEFAULT_PYTHON_VERSION = "3.8"
-SYSTEM_TEST_PYTHON_VERSIONS = ["3.7"]
-UNIT_TEST_PYTHON_VERSIONS = ["3.6", "3.7", "3.8"]
-
+DEFAULT_PYTHON_VERSION = "3.9"
+SYSTEM_TEST_PYTHON_VERSIONS = ["3.9"]
+UNIT_TEST_PYTHON_VERSIONS = ["3.6", "3.7", "3.8", "3.9"]
```
tseaver, line 32: This bit fights with autosynth

PR #256, opened 2020-11-12
chore: sort exceptions properly (manual synth)

tseaver:
I'm hoping that a manual synth will get the hashes in `synth.metadata` updated all at once, so that future "self" and "synthtoolls" autosynth runs can use the current latest version of googleapis.

- Restore path helper methods to generated clients.
- Enforce that 'request' argument to generated client methods is exclusive
  to flattened arguments.
- Surface 'transport' property for generated clients.

Closes #251
Closes #252


busunkim96 (approved):

tseaver:
@crwilcox, @czahedi  looks like @busunkim96 can't approve-for-merge here, either (`requiresCodeOwnerReviews: true` only allows members of @googleapis/firestore-dpe, looks like).

tseaver:
@crwilcox The main point of this PR is to do a full manual synth, trying to get around the fact that the three autosynth PRs can revert to older hashes for (especially, right now) `googleapis`.

The symptom of that reversion is that the `self` and `synthtools` builds are generating using `gapic-generator-python 0.33.2`, which has the bug of scrambling exception order (see #251, #253).

The rest of the changes (adding path helper methods, transport, etc.) are detailed in the commit description.

crwilcox (approved):

PR #263, opened 2020-11-27
docs: add documentation for documentsnapshot class

### LINKED ISSUE #262: Python-Firestore API Documentation
Hi, 

We are currently integrating Cloud Firestore (using the Python client) into one of our project. 

However we have some difficulties grasping all the capabilities because it seems the API documentation [here](https://googleapis.dev/python/firestore/latest/index.html) is incomplete.

I have to resort to browsing the GitHub source code to grasp which methods are available to perform a specific behavior.

Example: In the API Documentation there is no reference to `class DocumentSnapshot` I could find. 

Some of the questions, I was asking myself:
_How do I get a DocumentReference from a DocumentSnapshot instance ?_ 

I had to browse the `class DocumentSnapshot` source code and find that the property `reference` was available for that. 

Let me know if I'm not looking at the proper location. 
Thanks for your support.

HemangChothani:
Fixes #262 

- The document was there till the version `1.9.0` because the class was in the `document.py` file, but after that it was vanished as class moved to the `base_document.py` file.


docs/document.rst
```diff
@@ -4,3 +4,7 @@ Documents
 .. automodule:: google.cloud.firestore_v1.document
   :members:
   :show-inheritance:
+
+.. autoclass:: google.cloud.firestore_v1.base_document.DocumentSnapshot
+  :members:
+  :show-inheritance:
```
tseaver, line 10: We needed to document the whole `base_document` model, since done in PR #278.
```suggestion
```

tseaver (changes requested):

tseaver (approved):

crwilcox (approved):

Issue #268, opened 2020-12-09
Using client credentials with emulator


paulharter:
If I create a custom token with:

`https://www.googleapis.com/identitytoolkit/v3/relyingparty/verifyCustomToken?key=<key>`

and then generate creds with this:

`credentials.Credentials(token=token, scopes=_scopes)`

I can then use these credentials to connect with a Firestore database and the database rules will be respected. I don't use this in production but is very useful for testing. But this doesn't work when testing against the emulator rather than a real db. Testing with the emulator is great but I have some tests that require the db rules to work. I can do this with a subclass of `firestore.Client` but it would be nice to have this in library.

It can be done by changing the grpc.Channel used when connecting to the emulator so that it keeps the client._credentials rather than discarding them

I will submit a pull request for this shortly

Thanks

paulharter:
PR is here https://github.com/googleapis/python-firestore/pull/269

crwilcox:
merged feature.

PR #269, opened 2020-12-09
feat: support using client credentials with emulator

paulharter:
This changes the `grpc.Channel` used when connecting to a local emulator so that the client credentials are used. This allows for testing database rules against the emulator in python.

Fixes #https://github.com/googleapis/python-firestore/issues/268 

thanks

### DIFF
--- a/google/cloud/firestore_v1/base_client.py
+++ b/google/cloud/firestore_v1/base_client.py
@@ -174,8 +174,21 @@ class BaseClient(ClientWithProject):
 
         :return: grcp.Channel
         """
+        return grpc._channel.Channel(
+            self._emulator_host,
+            (),
+            self._local_composite_credentials()._credentials,
+            None,
+        )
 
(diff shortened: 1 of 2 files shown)


BenWhitehead:
This seems like a convenient thing to have, what do you think @crwilcox ?

crwilcox:
Thanks for the contribution @paulharter. In order for us to accept this we would need it to pass linting and coverage. The linting can be fixed by running `nox -s blacken` and verifying with `nox -s lint`. If you look at the kokoro build you will see what lines weren't covered, `google/cloud/firestore_v1/base_client.py                                                151      7     38      0    96%   178-199`

paulharter:
@crwilcox I've added a test and linted, I hope it's ok now.

Thank you for accepting this, it will make testing much easier.

crwilcox (approved):

crwilcox:
Thanks for the contribution @paulharter. This will go out in the next minor release of the library.

PR #290, opened 2021-01-15
tests: add system test to verify ArrayUnion upserts

### LINKED ISSUE #14: Firestore: using ArrayUnion in 'update()' or 'set(merge=true)' on nested data removes all previous data
#### Environment details

1. Cloud Firestore
2. Ubuntu 18.04
3. Python 3.7
4. google-cloud-firestore Version: 0.31.0

#### Steps to reproduce

  1. Use ArrayUnion([1,2,3]) on updating nested data in Document reference (see code example bellow)

#### Code example

```python
from google.cloud.firestore_v1beta1 import ArrayUnion

# creating reference
db = firestore.Client()
doc_ref = db.collection(u'test_collection').document(u'test_document')

# creating initial data with nested tree_1
doc_ref.set(
            {
                u'forest': {
                    u'tree_1': u'oak'
                }
            }
        )

# getting document after initial data
print('Here is initial data with 1 nested tree', doc_ref.get().to_dict())

# adding tree_2 to nested data (tree_1 remains in nested data)
doc_ref.set(
            {
                u'forest': {
                    u'tree_2': u'pine'
                }
            },
            merge=True
        )

print('Here is 2 trees in nested data after using set(merge=true)', doc_ref.get().to_dict())

# this code removes other nested data in document
# including tree_1 and tree_2, even if merge=true  is used
doc_ref.set(
            {
                u'forest': {
                    u'tree_3': ArrayUnion([u'spruce'])
                }
            },
            merge=True
        )

print('after using ArrayUnion, all previous nested data is removed', doc_ref.get().to_dict())
```

#### Stack trace
```
Here is initial data with 1 nested tree {'forest': {'tree_1': 'oak'}}
Here is 1 initial tree and 1 additional tree after set(merge=true) {'forest': {'tree_1': 'oak', 'tree_2': 'pine'}}
after using set(merge=true) with ArrayUnion all nested data is removed {'forest': {'tree_3': ['spruce']}}
```

Many Thanks!

craiglabenz:
Adds a system test demonstrating that ArrayUnion operation no longer deletes pre-existing data

Fixes #14 🦕

### DIFF
--- a/tests/system/test_system.py
+++ b/tests/system/test_system.py
@@ -1186,6 +1186,32 @@ def test_watch_query(client, cleanup):
         )
 
 
+def test_array_union(client, cleanup):
+    doc_ref = client.document("gcp-7523", "test-document")
+    cleanup(doc_ref.delete)
+    doc_ref.delete()
+    tree_1 = {"forest": {"tree-1": "oak"}}
+    tree_2 = {"forest": {"tree-2": "pine"}}
+    tree_3 = {"forest": {"tree-3": firestore.ArrayUnion(["spruce"])}}
+
+    doc_ref.set(tree_1)
+    expected = tree_1.copy()
+    assert doc_ref.get().to_dict() == expected
+
+    doc_ref.set(tree_2, merge=True)
+    expected["forest"]["tree-2"] = tree_2["forest"]["tree-2"]
+    assert doc_ref.get().to_dict() == expected
+
+    doc_ref.set(tree_3, merge=True)
+    expected["forest"]["tree-3"] = ["spruce"]
+    assert doc_ref.get().to_dict() == expected
+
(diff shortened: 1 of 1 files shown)


tseaver:
@craiglabenz I'm not sure that a system test is what #14 was asking for -- we were supposed to get a new "conformance test" (run as a data-driven unit test, rather than invoking the live back-end).  The advantage of such a test is that *all* language implementations are supposed to use the same set.

@BenWhitehead Can you please confirm whether a new conformance test is expected?

crwilcox:
@tseaver we are also going to add a conformance test, but that is bigger than just this client, and it makes sense to not block on ensuring this is protected. @craiglabenz and I are likely to pair on the conformance tests.

crwilcox (approved):

PR #316, opened 2021-02-12
refactor: removing usage of GetDocument proto in favor of BatchGetDocuments

craiglabenz:
Began refactoring away from GetDocument proto, which contains no `read_time` field and in general is a shortcut around `BatchGetDocuments`

This is pre-work necessary to support bundles.


google/cloud/firestore_v1/document.py
```diff
@@ -402,16 +403,26 @@ def get(
...
+            if self.id not in batched_response.missing:
+                data = _helpers.decode_dict(document.fields, self._client)
+                exists = True
+                create_time = document.create_time
+                update_time = document.update_time
+            else:
+                data = None
+                exists = False
+                create_time = None
+                update_time = None
```
crwilcox, line 419: does it make sense to gather these inits? I see we set create, update time to none twice for instance

google/cloud/firestore_v1/document.py
```diff
@@ -402,16 +403,26 @@ def get(
             create_time = None
             update_time = None
         else:
-            data = _helpers.decode_dict(document_pb.fields, self._client)
-            exists = True
-            create_time = document_pb.create_time
-            update_time = document_pb.update_time
+            batched_response: BatchGetDocumentsResponse = list(batch_response_pb)[0]
```
crwilcox, line 406: while this is not likely a problem, this could result in oob.

google/cloud/firestore_v1/document.py
```diff
@@ -389,11 +389,12 @@ def get(
...
                 its :attr:`exists` attribute will be ``False``.
         """
-        request, kwargs = self._prep_get(field_paths, transaction, retry, timeout)
+        request, kwargs = self._prep_batch_get(field_paths, transaction, retry, timeout)
 
         firestore_api = self._client._firestore_api
+        read_time: Timestamp = None
         try:
-            document_pb = firestore_api.get_document(
+            batch_response_pb: BatchGetDocumentsResponse = firestore_api.batch_get_documents(
```
crwilcox, line 397: a type seems possibly wrong. If this is a a BatchGetDocumentsResponse, why are we casting to a list and getting the first element on ln 406?

google/cloud/firestore_v1/document.py
```diff
@@ -402,16 +403,26 @@ def get(
...
         else:
-            data = _helpers.decode_dict(document_pb.fields, self._client)
-            exists = True
-            create_time = document_pb.create_time
-            update_time = document_pb.update_time
+            batched_response: BatchGetDocumentsResponse = list(batch_response_pb)[0]
+            read_time = batched_response.read_time
+            document: Document = batched_response.found
+            # TODO: Verify this check is correct.
+            if self.id not in batched_response.missing:
```
crwilcox, line 410: In the node sdk, `found` is used as the sentinel. Though missing ought to be the same? https://github.com/googleapis/nodejs-firestore/blob/1f8dc26032ef20ab874b70b1ae74988df484aec4/dev/src/index.ts#L1087

google/cloud/firestore_v1/document.py
craiglabenz: I was wondering about that. Will change.

google/cloud/firestore_v1/document.py
craiglabenz: What this is really highlighting is that that `NotFound` exception is no longer relevant, since missed documents still return empty responses. So, as far as I can tell, using the `BatchGetDocuments` proto, this list will always have an element.

But, all that said, I think I'll both remove the `NotFound` exception and explicitly check the length of this list.

google/cloud/firestore_v1/document.py
craiglabenz: Oh you're right - this is a StreamIterator or some such thing. Will change.

google/cloud/firestore_v1/document.py
craiglabenz: Here, `found` is the actual document, and `missing` is the full Firestore path to the document that didn't end up existing. So there are multiple ways to infer this 🤷

google/cloud/firestore_v1/document.py
craiglabenz: Truthiness check on `response` now avoids an empty list.

google/cloud/firestore_v1/document.py
```diff
@@ -389,29 +388,34 @@ def get(
...
+        exists = False
+        create_time = None
+        update_time = None
+
+        response_stream = firestore_api.batch_get_documents(
+            request=request, metadata=self._client._rpc_metadata, **kwargs,
+        )
+        response: List[BatchGetDocumentsResponse] = list(response_stream)
+
+        batch_pb: BatchGetDocumentsResponse = response[0]
```
crwilcox, line 405: nit: this used to actually be a pb type, do you think renaming to response or batch_response would be better?

google/cloud/firestore_v1/document.py
```diff
@@ -445,7 +449,9 @@ def collections(
             request=request, metadata=self._client._rpc_metadata, **kwargs,
         )
 
+        print("bout to iterate")
```
crwilcox, line 452: :)

google/cloud/firestore_v1/document.py
```diff
@@ -445,7 +449,9 @@ def collections(
...
         for collection_id in iterator:
+            print(f"collection_id: {collection_id}")
```
crwilcox, line 454: :)

tests/unit/v1/test_document.py
```diff
@@ -369,30 +369,35 @@ def _get_helper(
...
-        response.create_time = create_time
-        response.update_time = update_time
-
-        if not_found:
-            firestore_api.get_document.side_effect = NotFound("testing")
-        else:
-            firestore_api.get_document.return_value = response
+        read_time = 345
+        firestore_api = mock.Mock(spec=["batch_get_documents"])
+        # response = mock.create_autospec(document.Document)
```
crwilcox, line 383: remove?

crwilcox (changes requested):

google/cloud/firestore_v1/document.py
craiglabenz: I'm actually fairly started this was committed. I thought I made this change _after_ my last commit. This borderline feels like a glitch in the matrix...

google/cloud/firestore_v1/base_document.py
```diff
@@ -293,6 +293,32 @@ def _prep_get(
...
         return request, kwargs
 
+    def _prep_batch_get(
+        self,
+        field_paths: Iterable[str] = None,
+        transaction=None,
+        retry: retries.Retry = None,
+        timeout: float = None,
+    ) -> Tuple[dict, dict]:
+        """Shared setup for async/sync :meth:`get`."""
```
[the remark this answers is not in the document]
craiglabenz: The exposed function this supports is still `[async]document.get()`.

google/cloud/firestore_v1/document.py
```diff
@@ -389,29 +388,34 @@ def get(
...
+        create_time = None
+        update_time = None
+
+        response_stream = firestore_api.batch_get_documents(
+            request=request, metadata=self._client._rpc_metadata, **kwargs,
+        )
+        response: List[BatchGetDocumentsResponse] = list(response_stream)
+
+        batch_pb: BatchGetDocumentsResponse = response[0]
+        read_time = batch_pb.read_time
```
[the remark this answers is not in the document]
craiglabenz: Changed.

google/cloud/firestore_v1/document.py
craiglabenz: Updated to just `batch`.

google/cloud/firestore_v1/async_document.py
```diff
@@ -353,31 +357,34 @@ async def get(
                 :attr:`create_time` attributes will all be ``None`` and
                 its :attr:`exists` attribute will be ``False``.
         """
-        request, kwargs = self._prep_get(field_paths, transaction, retry, timeout)
+        from google.cloud.firestore_v1.base_client import _parse_batch_get
```
crwilcox, line 360: is this different than prep_batch_get?

craiglabenz: Yes, `prep_` is what sets up the request. `_parse` is what turns the response into a plain `DocumentSnapshot`.

google/cloud/firestore_v1/_helpers.py
```diff
@@ -1041,3 +1041,16 @@ def make_retry_timeout_kwargs(retry, timeout) -> dict:
         kwargs["timeout"] = timeout
 
     return kwargs
+
+
+class AsyncIter:
```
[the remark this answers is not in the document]
craiglabenz: Ah yes, it was originally going to be part of the real implementation, but that ended up only working in mocks. Reverted.

crwilcox (approved):

PR #319, opened 2021-03-02
feat: add firestore bundles

craiglabenz:


google/cloud/firestore_bundle/services/__init__.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
```
crwilcox, line 1: it seems the generated code is in this PR along with some manual edits. Would it be possible to separate the gen portions to shrink the size a bit?

craiglabenz: That's a great idea.

google/cloud/firestore_v1/_helpers.py
```diff
@@ -1041,3 +1084,126 @@ def make_retry_timeout_kwargs(retry, timeout) -> dict:
...
+        key: str = keys[0]
+
+        if key not in allowed_next_element_types:
+            raise ValueError(
+                f"Encountered BundleElement of type {key}. "
+                f"Expected one of {allowed_next_element_types}"
+            )
+
+        # Create and add our BundleElement
+        bundle_element: BundleElement = BundleElement.from_json(json.dumps(data))  # type: ignore
```
crwilcox, line 1149: Is this the only way to do this? it seems odd to me to serialize and deserialize again?

google/cloud/firestore_v1/base_query.py
```diff
@@ -396,6 +396,18 @@ def limit(self, count: int) -> "BaseQuery":
             all_descendants=self._all_descendants,
         )
 
+    @property
+    def limit_type(self) -> int:
+        """BundledQuery.LimitType equivalent of this query.
```
crwilcox, line 401: This seems a bit strange to me to end up in the query class, like the direction of dependence is almost reversed?

google/cloud/firestore_v1/_helpers.py
```diff
@@ -219,6 +232,36 @@ def encode_dict(values_dict) -> dict:
     return {key: encode_value(value) for key, value in values_dict.items()}
 
 
+def document_snapshot_to_protobuf(snapshot: "google.cloud.firestore_v1.base_document.DocumentSnapshot") -> Optional["google.cloud.firestore_v1.types.Document"]:  # type: ignore
```
crwilcox, line 235: Could this be added as an internal method on a DocumentSnapshot? ie `DocumentSnapshot._to_protobuf()`

google/cloud/firestore_v1/_helpers.py
```diff
@@ -219,6 +232,36 @@ def encode_dict(values_dict) -> dict:
...
+
+    return Document(
+        name=f"{snapshot._client._database_string}/{snapshot.reference.path}",
+        fields=encode_dict(snapshot._data),
+        create_time=snapshot.create_time,
+        update_time=snapshot.update_time,
+    )
+
+
+def parse_reference_value(reference_value: str) -> Dict[str, str]:
```
crwilcox, line 249: This looks to be extracting code from `reference_value_to_document` but that code hasn't been replaced to use this. Is this different from that code?  Perhaps this should give names to all of the splits? since it seems like for the other method we really only use the final split.

Is 'reference_value' a good name for this? it might be alright, though I think of these as 'document_path'. Curious about your thoughts on some of the terminology we use.

google/cloud/firestore_bundle/bundle.py
```diff
@@ -0,0 +1,395 @@
...
+            ] = bundle_element.document_metadata
+        elif type == "document":
+            parsed_path = _helpers.parse_reference_value(bundle_element.document.name)
+            snapshot = DocumentSnapshot(
+                data=_helpers.decode_dict(
+                    Document(mapping=bundle_element.document).fields, client
+                ),
+                exists=True,
+                reference=DocumentReference(
+                    parsed_path["collection_name"],
```
crwilcox, line 308: would it be helpful to make this a concrete type, or a named tuple (currently a dict)

crwilcox:
This generally looks pretty darn good! I have a few curiosities more than anything, wondering if there aren't places we can reduce some duplication :)

google/cloud/firestore_v1/base_query.py
craiglabenz: Hmm, yes I think you're right here. I had thought of this as limits obviously being generic to queries, but these limits are specifically namespaced within the BundledQuery proto.

Will reverse.

google/cloud/firestore_v1/_helpers.py
craiglabenz: It sure could. We may have miscommunicated, but I had understood one of our exchanges to mean this was the desired location 😆

google/cloud/firestore_bundle/bundle.py
craiglabenz: I originally shied away from a `NamedTuple` because I normally default to dataclasses (not 3.6-friendly), or attrs (didn't feel worth adding for this). But, a simple class/struct is a good compromise.

google/cloud/firestore_v1/_helpers.py
craiglabenz: I had expected it to be simpler, too, but this was all I found that worked. The form that seems like it would work, (`BundleElement(**data)`) errors out when it encounters Timestamps.

google/cloud/firestore_v1/_helpers.py
craiglabenz: Collapsed this functionality by introducing a utility class and deleting my original function. Ideally that utility class would be immutable, but wiring all of that up felt like a bit of overkill.

google/cloud/firestore_v1/_helpers.py
craiglabenz: PS - It is exposed in that way - the logic is just here (which is what I had understood us to settle on originally). Curious to hear whether that's the extent of what you meant, or if you're thinking of moving this entire block into that method.

google/cloud/firestore_bundle/bundle.py
```diff
@@ -0,0 +1,439 @@
...
+                method to get what this method expects.
+            ValueError: If the supplied name has already been added.
+        """
+        if not isinstance(query, BaseQuery):
+            raise ValueError(
+                "Attempted to add named query of type: "
+                f"{type(query).__name__}. Expected BaseQuery.",
+            )
+
+        if self.named_queries.get(name):
```
crwilcox, line 241: `name in self.named_queries`?

google/cloud/firestore_bundle/bundle.py
```diff
@@ -0,0 +1,439 @@
...
+            raise ValueError(f"Query name conflict: {name} has already been added.")
+
+        # Execute the query and save each resulting document
+        _read_time = self._save_documents_from_query(name, query)
+
+        # Actually save the query to our local object cache
+        self._save_named_query(name, query, _read_time)
+        # Flush this if it was cached
+        self._deserialized_metadata = None
```
crwilcox, line 250: these two lines show up often, but setting `_deserialized_metadata` doesn't map to 'flush from cache' in my head. Can you explain this? Perhaps this might be worth putting as a `_flush_cache` method (silly for one call but might allow for a good place to expand on how this works)

google/cloud/firestore_bundle/bundle.py
```diff
@@ -0,0 +1,439 @@
...
+        doc: DocumentSnapshot
+        for doc in query.stream():  # type: ignore
+            self.add_document(doc, query_name=name)
+            _read_time = doc.read_time
+        return _read_time
+
+    def _save_named_query(
+        self, name: str, query: BaseQuery, read_time: datetime.datetime,
+    ) -> None:
+        self.named_queries[name] = self._build_named_query(
```
crwilcox, line 271: since this is internal, any reason to not push the datetime conversion down into this?

google/cloud/firestore_v1/_helpers.py
crwilcox: I unfortunately do not recall the conversation :( I just found it unusual and worth asking :)

google/cloud/firestore_bundle/bundle.py
```diff
@@ -0,0 +1,439 @@
...
+    Generator,
+    List,
+    Optional,
+    Union,
+)
+
+
+class FirestoreBundle:
+    """A group of serialized documents and queries, suitable for
+    longterm storage.
```
schmidt-sebastian, line 48: Note that because of how resume tokens work in Firestore, it is cheaper to update a bundle more often. If a bundle is less than 30 minutes old, queries can be resumed cheaply on consuming clients.

"Longterm storage" is still a thing though, but we think that customers will frequently want to update their bundles.

google/cloud/firestore_bundle/bundle.py
```diff
@@ -0,0 +1,439 @@
...
+    def add(
+        self,
+        document_or_query_name: Union[DocumentSnapshot, str],
+        query: Optional[Query] = None,
+    ) -> "FirestoreBundle":
+        """Adds a document or entire query to the bundle.
+
+        This function is offered for convenience / API parity with other SDKs.
+        Most users will get better outcomes from using the more specific,
+        `add_document` or `add_named_query` methods.
```
schmidt-sebastian, line 99: I think this function is confusing and we can drop it for the better alternatives.

google/cloud/firestore_bundle/bundle.py
```diff
@@ -0,0 +1,439 @@
...
+    def add_document(
+        self, snapshot: DocumentSnapshot, query_name: Optional[str] = None,
+    ) -> "FirestoreBundle":
+        """Adds a document to the bundle.
+
+        Args:
+            snapshot (DocumentSnapshot): The fully-loaded Firestore document to
+                be preserved.
+            query_name (Optional[str]): If provided, also establishes a link
+                between the provided document and the referenced query.
```
schmidt-sebastian, line 155: Sorry I am doing a rush review here so I can't verify how everything works, but IMHO queries are atomic units and we can't add documents to them. The client will not surface the document if it doesn't match.

google/cloud/firestore_bundle/bundle.py
```diff
@@ -0,0 +1,439 @@
...
+                version=FirestoreBundle.BUNDLE_SCHEMA_VERSION,
+                total_documents=document_count,
+                total_bytes=len(buffer.encode("utf-8")),
+            )
+        )
+        return f"{self._compile_bundle_element(metadata)}{buffer}"
+
+    def _compile_bundle_element(self, bundle_element: BundleElement) -> str:
+        serialized_be: str = json.dumps(BundleElement.to_dict(bundle_element))
+        return f"{len(serialized_be)}{serialized_be}"
```
schmidt-sebastian, line 414: Does `len` return the length in bytes? Or is it the length of the JSON in String form? It should be bytes.

google/cloud/firestore_bundle/bundle.py
```diff
@@ -0,0 +1,439 @@
...
+        bundled_doc: "_BundledDocument"
+        for bundled_doc in self.documents.values():
+            if query_name and query_name not in bundled_doc.metadata.queries:
+                continue
+            yield bundled_doc.snapshot
+
+    def get_document(self, document_id: str) -> Optional[DocumentSnapshot]:
+        bundled_doc = self.documents.get(document_id)
+        if bundled_doc:
+            return bundled_doc.snapshot
```
schmidt-sebastian, line 428: I am getting somewhat hesitant about these methods. I don't see a  clear use case for them (other than our unit tests) and they don't exist on other Server or our Mobile SDKs, which creates API discrepancies and more work in the long run because one or two user will eventually ask us to add these APIs everywhere.

schmidt-sebastian (commented):
Some comments. Before merging, can you coordinate a test with wuandy@ to make sure that these Bundles load in our clients.

google/cloud/firestore_bundle/bundle.py
wu-hui: +1

google/cloud/firestore_bundle/bundle.py
wu-hui: Yes. The mapping from document to matched query name should be deducted from `add_named_query`, we should not expect clients to manually provide these mappings, if that is the intention here.

google/cloud/firestore_bundle/bundle.py
```diff
@@ -0,0 +1,439 @@
...
+        if query_name:
+            bundled_document = self.documents.get(full_document_path)
+            bundled_document.metadata.queries.append(query_name)  # type: ignore
+
+        self._update_last_read_time(snapshot.read_time)
+        # Flush this if it was cached
+        self._deserialized_metadata = None
+        return self
+
+    def add_named_query(self, name: str, query: BaseQuery) -> "FirestoreBundle":
```
wu-hui, line 208: Other SDKs take QuerySnapshot instead of Query. We made this choice to avoid having to surface errors properly when building the bundle, if the query run to error.

If I read this correctly, the query here will be run by the bundle builder? Please use snapshot instead, for above reason and for consistency.

google/cloud/firestore_bundle/bundle.py
```diff
@@ -0,0 +1,439 @@
...
+        bundle.add(
+            'top-ten-hamburgers',
+            db.collection('hamburgers').limit(limit=10)._query(),
+        )
+        serialized: str = bundle.build()
+
+        # Store somewhere like your GCS or your device's file system
+
+        # Some time later after loading `serialized` again
+        bundle = _helpers.deserialize_bundle(serialized)
```
wu-hui, line 71: If I recall our offline discussion correctly, `deserialize_bundle` is a helper for testing for now, please make it equivalent
 to Java's `@TestOnly` and `@Hide`, and please do not mention it in public doc/comment.

google/cloud/firestore_bundle/bundle.py
wu-hui: my quick stack overflow search suggests `len(s.encode('utf-8'))`..

google/cloud/firestore_bundle/bundle.py
wu-hui: +1

tests/unit/v1/test_bundle.py
```diff
@@ -0,0 +1,468 @@
...
+        """
+        client = self.get_client()
+        template = client._database_string + "/documents/col/{}"
+        document_ids = document_ids or ["doc-1", "doc-2"]
+        documents = [
+            RunQueryResponse(
+                transaction=b"",
+                document=Document(
+                    name=template.format(document_id),
+                    fields=_helpers.encode_dict({"hello": "world"}),
```
wu-hui, line 72: Can we add some more data types, also include some non-ASCII chars?

wu-hui (changes requested):

google/cloud/firestore_bundle/bundle.py
craiglabenz: Oh I see. I was mimicking the function signatures from Java, but maybe I misunderstood their intent. I'll make this change.

google/cloud/firestore_bundle/bundle.py
craiglabenz: I had that same suspicion and used that first, and even included compound emojis, but it never seemed to return a different answer. But, I'll revisit this.

google/cloud/firestore_bundle/bundle.py
craiglabenz: What is the intended way that developers would access the content within a Bundle? It feels like developers will either need to exist or will write this same code on their own.

Am I missing anything there?

google/cloud/firestore_bundle/bundle.py
craiglabenz: Sadly, the Python SDK has no concept of a `QuerySnapshot` 😭 . I very much wanted to use that, instead.

tests/unit/v1/test_bundle.py
craiglabenz: Ah yes, I tested these manually, but there should be unit tests for them as well. I will add some.

google/cloud/firestore_bundle/bundle.py
craiglabenz: I just double checked this and confirmed that `json.dumps` encodes non-ASCII characters.

google/cloud/firestore_v1/_helpers.py
craiglabenz: Resolving this since the class's `_to_protobuf()` exists and calls this method (which is a pattern that exists elsewhere in the library).

google/cloud/firestore_bundle/bundle.py
craiglabenz: It makes sense that documents updated intermittently, but read very frequently, would be updated regularly. I'd previously imagined this as mostly being used for lookup tables, and the like.

At any rate, I've expanded on the docstring's opening sentence.

google/cloud/firestore_bundle/bundle.py
craiglabenz: Removed mention, and relocated more functions to `_helpers`.

google/cloud/firestore_bundle/bundle.py
craiglabenz: I'm happy to drop this - I'd only reluctantly added it for parity (as the docstring says). But if you two don't think it's necessary, I am happy to nuke it.

google/cloud/firestore_bundle/bundle.py
craiglabenz: Moved to dedicated method with comment.

google/cloud/firestore_bundle/bundle.py
craiglabenz: Good idea. Change is made.

google/cloud/firestore_bundle/bundle.py
craiglabenz: Moved these functions to `_helpers`.

google/cloud/firestore_bundle/bundle.py
```diff
@@ -0,0 +1,363 @@
...
+        self.latest_read_time: Timestamp = Timestamp(seconds=0, nanos=0)
+        self._deserialized_metadata: Optional[BundledDocumentMetadata] = None
+
+    def add_document(self, snapshot: DocumentSnapshot) -> "FirestoreBundle":
+        """Adds a document to the bundle.
+
+        Args:
+            snapshot (DocumentSnapshot): The fully-loaded Firestore document to
+                be preserved.
+            query_name (Optional[str]): If provided, also establishes a link
```
wu-hui, line 87: Delete this line.

google/cloud/firestore_bundle/bundle.py
wu-hui: Sorry, I have no idea `QuerySnapshot` is not a thing in Python...I am surprised.

However, we could potentially accept whatever `Query.get` returns here (looks like it returns a list of `DocumentSnapshot`)?

The benefit, as stated in the original comment, is you do not need to worry about reporting error when `build()`. Otherwise, we should try to tell the users exactly which named query failed to fetch when we `build()`, we also need to decide if one of the query failed, should `build()` still proceed.

wu-hui (commented):
Thanks for incorporating the comments! I think we are close now.

I do want to discuss the possibility of accepting list of documents for `addNamedQuery`, please let me know what do you think.

google/cloud/firestore_bundle/bundle.py
craiglabenz: Good catch!

craiglabenz:
That's reasonable, @wu-hui. Adding now.

craiglabenz:
@wu-hui On second thought, accepting a list of `DocumentSnapshot` instances creates a new unsolved problem around how to reconstruct a proper `NamedQuery` object. I might vote that we defer that option for now.

schmidt-sebastian:
@wu-hui Can we just treat queries and documents as independent? The Python SDK could offer APIs to add named queries and documents, but we don't necessarily need to associate the documents with the queries. The Mobile clients should still be able to handle this data (correct me if I am wrong).

I'm also very surprised that Python still does not have QuerySnapshots and a bit scared to find out what this means for the Watch API :)

wu-hui:
> @wu-hui Can we just treat queries and documents as independent? The Python SDK could offer APIs to add named queries and documents, but we don't necessarily need to associate the documents with the queries. The Mobile clients should still be able to handle this data (correct me if I am wrong).
> 
> I'm also very surprised that Python still does not have QuerySnapshots and a bit scared to find out what this means for the Watch API :)

I think we might have to accept `Query` and run on behalf of user for Python, there is no other way to get `StructuredQuery` otherwise.  WDYT?

google/cloud/firestore_bundle/bundle.py
```diff
@@ -0,0 +1,363 @@
...
+
+    Usage:
+
+        from google.cloud.firestore import Client
+        from google.cloud.firestore_bundle import FirestoreBundle
+        from google.cloud.firestore import _helpers
+
+        db = Client()
+        bundle = FirestoreBundle('my-bundle')
+        bundle.add('all-users', db.collection('users')._query())
```
wu-hui, line 60: Should be `add_named_query`. There are other places where update is needed.

google/cloud/firestore_bundle/bundle.py
```diff
@@ -0,0 +1,363 @@
...
+
+        Example:
+
+            from google.cloud import firestore
+
+            db = firestore.Client()
+            collection_ref = db.collection(u'users')
+
+            bundle = firestore.FirestoreBundle('my bundle')
+            bundle.add('app-users', collection_ref._query())
```
wu-hui, line 302: `add_named_query` instead.

google/cloud/firestore_bundle/bundle.py
```diff
@@ -0,0 +1,363 @@
...
+                create_time=_helpers.build_timestamp(),
+                version=FirestoreBundle.BUNDLE_SCHEMA_VERSION,
+                total_documents=document_count,
+                total_bytes=len(buffer.encode("utf-8")),
+            )
+        )
+        return f"{self._compile_bundle_element(metadata)}{buffer}"
+
+    def _compile_bundle_element(self, bundle_element: BundleElement) -> str:
+        # serialized_be: str = json_format.MessageToJson(BundleElement.to_dict(bundle_element))
```
wu-hui, line 345: Delete this line.

wu-hui (approved):
LGTM with some nits on comments

wu-hui:
LGTM!

crwilcox (approved):

PR #325, opened 2021-03-12
feat: adds synthed bundle protos

craiglabenz:
Substring of #319, landing first to reduce noise there


crwilcox (approved):

crwilcox:
Coverage is low due to gen code, though #319 will bring it up. Going to allow the drop.

PR #340, opened 2021-04-23
fix: Allow underspecifying a cursor

### LINKED ISSUE #339: Library raises when fewer cursors than order by clauses
Firestore supports setting multiple order by clauses, but then only using one of them for a cursor. 
https://cloud.google.com/firestore/docs/query-data/query-cursors#set_cursor_based_on_multiple_fields

```
    def test__normalize_cursor_as_dict_extra_orders_ok(self):
        cursor = ({'name': 'Springfield'}, True)
        query = self._make_one(mock.sentinel.parent).order_by("name").order_by("state")

        query._normalize_cursor(cursor, query._orders)
```

also, I'd expect unnamed cursors to work, as they do in java for instance

```
 def test__normalize_cursor_extra_orders_ok(self):
        cursor = (['Springfield'], True)
        query = self._make_one(mock.sentinel.parent).order_by("name").order_by("state")

        query._normalize_cursor(cursor, query._orders)
```

Today, this raises the following error:
```
ValueError: The "order by" field path 'state' is not present in the cursor data {'name': 'Springfield'}. All fields sent to order_by() must be present in the fields if passed to one of start_at() / start_after() / end_before() / end_at() to define a cursor.
```

Open Question: Is this against the rules (cursor on something not ordered?
```
cursor = ({"a": 1}, True)
query = self._make_one(mock.sentinel.parent).order_by("b", "ASCENDING")
```

crwilcox:
Draft to help show where Python differs w.r.t cursors, orderby

Fixes #339 🦕


craiglabenz (approved):
LGTM

google/cloud/firestore_v1/base_query.py
```diff
@@ -756,9 +760,10 @@ def _normalize_cursor(self, cursor, orders) -> Optional[Tuple[Any, Any]]:
                 except KeyError:
                     msg = _MISSING_ORDER_BY.format(order_key, data)
                     raise ValueError(msg)
+
             document_fields = values
 
-        if len(document_fields) != len(orders):
+        if len(document_fields) > len(orders):
```
schmidt-sebastian, line 766: Nit: Throw this first to skip the work above

schmidt-sebastian (approved):

google/cloud/firestore_v1/base_query.py
crwilcox: I wanted to do that but it would require a bit more restructuring (documentsnapshots get converted to dicts to share this code), this is then used for both cases there and if it was an array.

PR #352, opened 2021-05-10
chore: migrate to owl bot

parthea:
This PR migrates from autosynth to [owl bot](https://github.com/googleapis/repo-automation-bots/tree/master/packages/owl-bot).  owl bot will save time for maintainers as it will automatically open PRs when there are updates in [googleapis-gen](https://github.com/googleapis/googleapis-gen/tree/master/google) without requiring maintainers to run `synthtool` to build the client from protos. Additionally, similar to autosynth, PRs will be automatically opened when there are template updates.

PR #365, opened 2021-06-06
chore: update owlbot.py to properly copy folders from googleapis-gen

parthea:
`owlbot.py` was not configured correctly in #352 as a result, changes were not pull in from googleapis-gen. This PR includes the following fixes/features:

chore: upgrade gapic-generator-python to 0.46.3
fix: add async client to %name_%version/init.py
chore: add autogenerated snippets
chore: remove auth, policy, and options from the reserved names list
feat: support self-signed JWT flow for service accounts
chore: enable GAPIC metadata generation
chore: sort subpackages in %namespace/%name/init.py


owlbot.py
```diff
@@ -115,20 +137,16 @@ def update_fixup_scripts(library):
...
-s.move(
-    templated_files,
-)
-
-s.replace(
-    "noxfile.py",
-    "GOOGLE_APPLICATION_CREDENTIALS",
-    "FIRESTORE_APPLICATION_CREDENTIALS",
-)
+s.move(templated_files)
```
tseaver, line 140: The change to replace `GOOGLE_APPLICATION_CREDENTIALS` with `FIRESTORE_APPLICATION_CREDENTIALS` must be preserved:  firestore system tests rely on a different environment variable.

owlbot.py
```diff
@@ -13,92 +13,114 @@
...
-            library / f"google/cloud/bundle",
-            f"google/cloud/firestore_bundle",
-        )
-        s.move(library / f"tests", f"tests")
+    s.replace(
+        library / "google/cloud/bundle/__init__.py",
+        "from .types.bundle import BundleMetadata\n"
+        "from .types.bundle import NamedQuery\n",
+        "from .types.bundle import BundleMetadata\n"
+        "from .types.bundle import NamedQuery\n\nfrom .bundle import FirestoreBundle\n",
```
tseaver, line 110: For clarity, either break across lines:
```suggestion
        "from .types.bundle import NamedQuery\n"
        "\n"
        "from .bundle import FirestoreBundle\n",
```
or else use triple-quoted strings.

tseaver (changes requested):

owlbot.py
tseaver:
```suggestion
s.move(templated_files)

s.replace(
    "noxfile.py",
    "GOOGLE_APPLICATION_CREDENTIALS",
    "FIRESTORE_APPLICATION_CREDENTIALS",
)
```

tseaver: I guess we've decided to drop the check for the environment variable in `noxfile.py` -- I'm not sure why, but it is gone in (some) other libraries, at least.

tseaver (approved):

PR #370, opened 2021-06-15
fix: update type hint for async_collection.document() to AsyncDocumen…

### LINKED ISSUE #369: Change Type Hit at AsyncCollectionReference.document

Return Type Hint at AsyncCollectionReference.document is always DocumentReference, but actually AsyncDocumentReference.
My Editor(Jetbrains IDE) tells me that AsyncDocumentReference is wrong Type.
スクリーンショット 2021-06-15 17 54 31

I think to change Return Type Hint is the best solution:

```python
def document(self, document_id: str = None) -> AsyncDocumentReference:
    """..."""
    return super().document(document_id)
```

 **Describe alternatives you've considered**

I think this solution can change Return Type Hint simply.

 **Additional context**

Nothing.

kolea2:
…tReference

Fixes #369 🦕


google/cloud/firestore_v1/async_collection.py
```diff
@@ -112,6 +112,23 @@ async def add(
...
+            document_id (Optional[str]): The document identifier
+                within the current collection. If not provided, will default
+                to a random 20 character string composed of digits,
+                uppercase and lowercase and letters.
+
+        Returns:
+            :class:`~google.cloud.firestore_v1.document.async_document.AsyncDocumentReference`:
+            The child document.
+        """
+        return super(AsyncCollectionReference, self).document(document_id)
```
craiglabenz, line 130: Hmm, it's unfortunate that we still use the Python 2 syntax here, though I see that's the pattern across the library and you were just following it. Nothing to do here except collectively shrug our shoulders.

craiglabenz (approved):
LGTM

google/cloud/firestore_v1/async_collection.py
crwilcox: You are welcome to go through and refactor this. This is less a choice and more a fact that this library evolved from Py2/3 compliant code and we didn't refactor more than necessary to use new features really (AsyncIO, f strings, etc)

PR #375, opened 2021-06-16
feat: added `write` method to batch classes

craiglabenz:
Adds a new `BulkWriteBatch` class to accompany the existing `WriteBatch` and `AsyncWriteBatch` classes to use the new and preferred RPC, `BatchWriteRequest`. `BulkWriteBatch` is designed to be used with `BulkWriter`, which is opaquely async, and thus there is no companion `AsyncBulkWriteBatch` class.

This is a component of the BulkWriter feature, which will ultimately consume arbitrarily large data feeds, chunk them in to BulkWriteBatch instances, and use the `BatchWriteRequest` PRC to submit each write operation to the server.


google/cloud/firestore_v1/base_batch.py
```diff
@@ -33,14 +33,22 @@ class BaseWriteBatch(object):
...
+        write_ctx (bool):
+            Controls whether this instance should call `commit` or `write` upon
+            exiting as a context manager. This parameter has no impact if you
+            do not use the instance as a context manager.
     """
 
-    def __init__(self, client) -> None:
+    def __init__(self, client, write_ctx: bool = False) -> None:
         self._client = client
+        self._write_ctx = write_ctx
```
craiglabenz, line 44: This variable is one of the only interesting parts of this PR. For context, the `[Async]WriteBatch` class can be used as a context manager, and upon exiting such a state, called its most interesting method: `commit`. 

However, this PR adds a sibling method, `write`, which uses a new protobuf object underneath. Thus, how an `[Async]WriteBatch` instance should behave upon exiting its context manager form is no longer clear. The moment of transforming an instance into a context manager does not offer any levers for control, so I opted to add a flag to the constructor.

Open to ideas, but that's the situation :)

tseaver: The other option would be to subclass `BaseWriteBatch` as `BaseBulkWriteBatch` (and the non-base versions), overriding `__exit__` / `__aexit__` to call `self.write()`.  ISTM that might be a clearer signal.  Your split of the testcase class below seems to signal that, too.

google/cloud/firestore_v1/base_batch.py
```diff
@@ -33,14 +33,22 @@ class BaseWriteBatch(object):
...
         self._write_pbs = []
         self.write_results = None
         self.commit_time = None
 
+    def __len__(self):
+        return len(self._write_pbs)
+
```
tseaver, line 51: How do you imagine `len(batch)` being used?

google/cloud/firestore_v1/base_batch.py
craiglabenz: It wasn't obvious to me whether that was overkill for this one line branch, but maybe it is the right way to go. I'll give it a look.

google/cloud/firestore_v1/base_batch.py
craiglabenz: `BulkWriter` looks to cut off batch sizes at 20 operations, so this allows that code to evaluate `len(batch)` instead of `len(batch._write_pbs)`.

craiglabenz:
I am comfortable merging this into the release branch, @tseaver and @crwilcox.

google/cloud/firestore_v1/bulk_batch.py
```diff
@@ -0,0 +1,87 @@
...
+            write result contains an ``update_time`` field.
+        """
+        request, kwargs = self._prep_commit(retry, timeout)
+
+        save_response = self._client._firestore_api.batch_write(
+            request=request, metadata=self._client._rpc_metadata, **kwargs,
+        )
+
+        self._write_pbs = []
+        self.write_results = results = list(save_response.write_results)
```
crwilcox, line 75: Is this intermediate assignment to results to manage some sort of race? If so, is there not also an issue with commit prepping and setting `_write_pbs` to empty?

craiglabenz: I'm not aware of any possibly race conditions here, but if you think there is we should chat more.

Also worth noting that this functionality is more or less copied out of the older `batch.py` file.

PR #388, opened 2021-07-02
fix: reseed for each auto id on 3.6 to avoid collisions

### LINKED ISSUE #346: New document ID not unique after fork
Originally reported here: https://github.com/firebase/firebase-admin-python/issues/548

If firebase_admin is imported before os.fork(), auto-generated document IDs are repeated causing data to be overwritten.

#### Environment details

  - OS type and version: Linux Mint 20
  - Python version: Python 3.6.13
  - pip version: pip 21.1
  - `google-cloud-firestore` version: 2.1.0

#### Steps to reproduce

Import firebase_admin, fork, create firestore client, write a new document.

#### Code example

```python
#!/usr/bin/env python3.6

import os
import firebase_admin
from firebase_admin import credentials
from firebase_admin import firestore

pid = os.fork()

cred = credentials.Certificate('service_account.json')
firebase_admin.initialize_app(cred)

client = firestore.client()

if pid == 0:
    reference = client.collection('Test').document()
    reference.set({'a': 'b'})

    print('Child: ' + reference.id)
else:
    reference = client.collection('Test').document()
    reference.set({'a': 'c'})

    print('Parent: ' + reference.id)
```
Output:
```
Child: TCsit2S0VQgvqm0aHNTc
Parent: TCsit2S0VQgvqm0aHNTc
```

crwilcox:
Fixes #346 🦕

Upon further investigation, this is only an issue for python 3.6. In 2022, when 3.6 is no longer supported, this patch could be removed.

### DIFF
--- a/google/cloud/firestore_v1/base_collection.py
+++ b/google/cloud/firestore_v1/base_collection.py
@@ -15,7 +15,6 @@
 """Classes for representing collections for the Google Cloud Firestore API."""
 import random
 import sys
-import uuid
 
 from google.api_core import retry as retries  # type: ignore
 
@@ -459,10 +458,9 @@ def _auto_id() -> str:
     """
     if sys.version_info < (3, 7):
         # TODO: remove when 3.6 support is discontinued.
-        # On python 3.6, random will provide the same results when forked. Reseed using a
-        # uuid4 on each iteration to avoid collisions.
-        seed = str(uuid.uuid4())
-        random.seed(seed)
+        # On python 3.6, random will provide the same results when forked. Reseed 
+        # on each iteration to avoid collisions.
+        random.seed()
 
     return "".join(random.choice(_AUTO_ID_CHARS) for _ in range(20))
 
--- a/google/cloud/firestore_v1/base_collection.py
+++ b/google/cloud/firestore_v1/base_collection.py
@@ -458,7 +458,7 @@ def _auto_id() -> str:
     """
     if sys.version_info < (3, 7):
         # TODO: remove when 3.6 support is discontinued.
-        # On python 3.6, random will provide the same results when forked. Reseed 
+        # On python 3.6, random will provide the same results when forked. Reseed
         # on each iteration to avoid collisions.
         random.seed()
 
--- a/google/cloud/firestore_v1/base_collection.py
+++ b/google/cloud/firestore_v1/base_collection.py
@@ -14,6 +14,7 @@
 
 """Classes for representing collections for the Google Cloud Firestore API."""
 import random
+import sys
 import uuid
 
 from google.api_core import retry as retries  # type: ignore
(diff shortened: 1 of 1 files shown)


google/cloud/firestore_v1/base_collection.py
```diff
@@ -455,6 +457,13 @@ def _auto_id() -> str:
         str: A 20 character string composed of digits, uppercase and
         lowercase and letters.
     """
+    if sys.version_info < (3, 7):
+        # TODO: remove when 3.6 support is discontinued.
+        # On python 3.6, random will provide the same results when forked. Reseed using a
+        # uuid4 on each iteration to avoid collisions.
+        seed = str(uuid.uuid4())
+        random.seed(seed)
```
tseaver, line 465: `uuid.uuid4` itself uses `os.urandom` module as its source of entropy, which is exactly what `random.seed` does if the `seed` argument is `None`, this would be simpler as:

```suggestion
        random.seed()
```

(and remove the import and use of `uuid`).

tseaver (changes requested):

google/cloud/firestore_v1/base_collection.py
crwilcox: I tried to reproduce a collision with just `random.seed()` and did not, but I have conerns as the docs state this will use system time as a seed, which seems to me there could be a collision.

Running the below code:

```
if pid == 0:
    r = random.random()
    id = uuid.uuid4()
    reference = client.collection("Test").document()
    print("Child: random.random:", r, "uuid4:", id, "documentId:", reference.id)
else:
    r = random.random()
    id = uuid.uuid4()
    reference = client.collection("Test").document()
    print("Parent: random.random:", r, "uuid4:", id, "documentId:", reference.id)
```

```
Parent: random.random: 0.28313230731702665 uuid4: 738ffed5-c52b-44cf-ba9b-1118ad9921c0
Child: random.random: 0.28313230731702665 uuid4: 0ae5a7ae-4345-4154-8c9e-c988bda1cd7e
```

Across forks `random.random` is the same, but `uuid4` is unique. The source in cpython for uuid4 is identical in 3.6 and 3.9 branches, but random.py has seen a reasonable amount of change, possibly in the sha512 c module also.

It does seem that `random.seed` improves things, I am just not sure it won't ever collide.

tseaver: `random.seed` prefers to use the system `urandom` value, but falls back to system time + PID if it the `urandom` bits are exhausted, or the system doesn't support `urandom`.  This behavior is consistent across 3.6 - 3.9.

crwilcox: Seems like we can simplify then. I am not going to spend too much time seeing as this fix will be redundant in 6 months as 3.7+ behave correctly. There seems to be a small change between 3.6 and 3.7 with a lack of randomness on fork.

craiglabenz (approved):
Pseudorandom? More like SUDOrandom, amirite?

google/cloud/firestore_v1/base_collection.py
tseaver: LGTM

tseaver (approved):

PR #392, opened 2021-07-14
fix: expect 'DocumentSnapshot.__hash__' to expect 'DatetimeWithNanosecons'

### LINKED ISSUE #391: DocumentSnapshots cannot be stored in a set, unlike DocumentReference
#### Environment details

  - OS type and version: 64-bit and Ubuntu 18.04.5 LTS
  - Python version: `python 3.7`
  - pip version: `pip 21.1.3`
  - google-cloud-firestore version: `google-cloud-firestore 2.1.0`

#### Steps to reproduce

  1. Add some documents to Firestore and store the document snapshots in a variable ( say `result` )
  2. Form a set of the documents while fetching them `set(result)`

#### Code example

```python
fs_client = firestore.Client()
result=set()
query = fs_client.collection_group(<collection-group-name>)    
for document_snapshot in query.stream():
        result.add(document_snapshot)
```

#### Stack trace
```
  File "<folder>/venv/lib/python3.7/site-packages/google/cloud/firestore_v1/base_document.py", line 372, in __hash__
    seconds = int(self.update_time.timestamp())
AttributeError: timestamp
```

The same does not throw an error when we store `DocumentReference` objects in a set.
Also, modifying `__hash__` as below works:
```python
    def __hash__(self):
        seconds = int(self.update_time.seconds)
        nanos = self.update_time.nanos
        return hash(self._reference) + hash(seconds) + hash(nanos)
```

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

Arushi201296:
Fixes #391  🦕

### DIFF
--- a/google/cloud/firestore_v1/base_document.py
+++ b/google/cloud/firestore_v1/base_document.py
@@ -368,8 +368,8 @@ class DocumentSnapshot(object):
         return self._reference == other._reference and self._data == other._data
 
     def __hash__(self):
-        seconds = int(self.update_time.timestamp())
-        nanos = self.update_time.nanosecond
+        seconds = int(self.update_time.seconds) 
+        nanos = self.update_time.nanos
         return hash(self._reference) + hash(seconds) + hash(nanos)
 
     @property


tseaver (changes requested):
Thanks very much for the patch!  The PR needs to make corresponding changes to the unit test(s) for `__hash__`, found in `tests/unit/v1/test_base_document.py`.

You can run the tests locally with `nox`:  install it via `pip install nox` and then run `nox -s unit-3.8`.

tseaver:
Tagged as `do not merge` pending resolution of the discussion in #398.

crwilcox:
Note that there are current test failures to address 
image

tseaver:
@Arushi201296 I've added the needed test in #467, and so I'm closing this PR.  Again, thanks for the report and the patch!

PR #397, opened 2021-07-15
feat: added new batch class for BulkWriter

craiglabenz:
As a component of the BulkWriter feature, this PR:

* Adds a new `BulkWriteBatch` class whose `commit` method uses the BatchWriteRequest proto.
* Adds `__len__` and `__contains__` magic methods to all Batch classes for interrogation of which write operations they contain at any given moment
* Adds the `abc.ABCMeta` metaclass to enforce that all batches implement `commit`

### DIFF
--- a/google/cloud/firestore_v1/base_batch.py
+++ b/google/cloud/firestore_v1/base_batch.py
@@ -14,16 +14,16 @@
 
 """Helpers for batch requests to the Google Cloud Firestore API."""
 
-
-from google.cloud.firestore_v1 import _helpers
+import abc
+from typing import Dict, Union
 
 # Types needed only for Type Hints
-from google.cloud.firestore_v1.document import DocumentReference
+from google.api_core import retry as retries  # type: ignore
+from google.cloud.firestore_v1 import _helpers
+from google.cloud.firestore_v1.base_document import BaseDocumentReference
 
-from typing import Union
 
-
-class BaseWriteBatch(object):
+class BaseBatch(metaclass=abc.ABCMeta):
     """Accumulate write operations to be sent in a batch.
 
     This has the same set of methods for write operations that
@@ -38,8 +38,15 @@ class BaseWriteBatch(object):
     def __init__(self, client) -> None:
         self._client = client
         self._write_pbs = []
+        self._document_references: Dict[str, BaseDocumentReference] = {}
         self.write_results = None
         self.commit_time = None
+
+    def __len__(self):
+        return len(self._document_references)
+
+    def __contains__(self, reference: BaseDocumentReference):
+        return reference._document_path in self._document_references
 
     def _add_write_pbs(self, write_pbs: list) -> None:
         """Add `Write`` protobufs to this transaction.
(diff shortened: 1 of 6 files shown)


craiglabenz:
This one is pretty much ready to go. It introduces a new class (`BulkWriteBatch`) that is extremely similar to a previously existing class (`WriteBatch`), with mimicked test classes. Once this is in the release branch, the rest of the BulkWriter implementation will be the actual BulkWriter class itself 🎉

google/cloud/firestore_v1/bulk_batch.py
```diff
@@ -0,0 +1,89 @@
...
+from google.cloud.firestore_v1.base_batch import BaseBatch
+from google.cloud.firestore_v1.types.firestore import BatchWriteResponse
+
+
+class BulkWriteBatch(BaseBatch):
+    """Accumulate write operations to be sent in a batch. Use this over
+    `WriteBatch` for higher volumes (e.g., via `BulkWriter`) and when the order
+    of operations within a given batch is unimportant.
+
+    Because the order in which individual write operations land in the database
```
crwilcox, line 29: nit: should this read "are applied to" instead of "land in". Wonder if that is colloquial.

google/cloud/firestore_v1/base_batch.py
```diff
@@ -146,9 +162,15 @@ def delete(
                 state of the document before applying changes.
         """
         write_pb = _helpers.pb_for_delete(reference._document_path, option)
+        self._document_references[reference._document_path] = reference
         self._add_write_pbs([write_pb])
 
-    def _prep_commit(self, retry, timeout):
```
crwilcox, line 151: it seems like this is still used, is there a reason to remove it from the base?

google/cloud/firestore_v1/bulk_batch.py
```diff
@@ -0,0 +1,89 @@
...
+from google.api_core import retry as retries  # type: ignore
+
+from google.cloud.firestore_v1 import _helpers
+from google.cloud.firestore_v1.base_batch import BaseBatch
+from google.cloud.firestore_v1.types.firestore import BatchWriteResponse
+
+
+class BulkWriteBatch(BaseBatch):
+    """Accumulate write operations to be sent in a batch. Use this over
+    `WriteBatch` for higher volumes (e.g., via `BulkWriter`) and when the order
```
crwilcox, line 26: We specify higher and lower volumes, but these are relative. is there an (approximate) threshold that you would choose one over the other?

craiglabenz: Probably in the tens of thousands, but I don't know the performance / throughput numbers for the previous bulk patterns that this pseudo-replaces. At any rate, I think that information is probably best left for docs / samples.

google/cloud/firestore_v1/base_batch.py
craiglabenz: It's now moved to `BaseWriteBatch`, which is the base class of the pre-existing `[Async]WriteBatch` class. This negative delta is in `BaseBatch`, which handles logic shareable between `[Async]WriteBatch` and `BulkBatch`.

PR #406, opened 2021-07-27
refactor: added BaseQuery._copy method

craiglabenz:
Consolidates query copying through one choke point which can be used to ensure any/all future parameters are also copied in all places

### DIFF
--- a/google/cloud/firestore_v1/_helpers.py
+++ b/google/cloud/firestore_v1/_helpers.py
@@ -76,6 +76,7 @@ class EmptyParam:
     
     When seen, the pseudo-singleton `empty_param` value below serves as a sentinel
     to indicate that the parameter was ignored by calling code."""
+
     def __eq__(self, other):
         return isinstance(other, EmptyParam)
 
--- a/google/cloud/firestore_v1/_helpers.py
+++ b/google/cloud/firestore_v1/_helpers.py
@@ -70,20 +70,6 @@ _GRPC_ERROR_MAPPING = {
 }
 
 
-class EmptyParam:
-    """Default value for optional parameters used to disambiguate between
-    calling code passing `None` versus ignoring the parameter entirely.
-    
-    When seen, the pseudo-singleton `empty_param` value below serves as a sentinel
-    to indicate that the parameter was ignored by calling code."""
-
-    def __eq__(self, other):
-        return isinstance(other, EmptyParam)
-
-
-empty_param = EmptyParam()
-
-
 class GeoPoint(object):
     """Simple container for a geo point value.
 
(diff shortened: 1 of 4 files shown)


google/cloud/firestore_v1/_helpers.py
```diff
@@ -70,6 +70,20 @@
 }
 
 
+class EmptyParam:
```
chrisrossi, line 73: This is a little more trouble than I usually go to for this. Generally something like:

_not_passed = object()

def f(x, y=_not_passed):
    if y is _not_passed:
        etc...

craiglabenz: Oh yeah, that's less lines :)

tests/unit/v1/test_async_collection.py
```diff
@@ -51,6 +51,9 @@ def test_query_method_matching(self):
         from google.cloud.firestore_v1.async_query import AsyncQuery
 
         query_methods = self._get_public_methods(AsyncQuery)
+        # Remove the one query-only method
```
chrisrossi, line 54: Maybe _copy is a private method? Is the intent for purely internal use or do you foresee users calling it?

craiglabenz: I suppose it could be private 🤔

chrisrossi (approved):
LGTM

google/cloud/firestore_v1/base_query.py
```diff
@@ -244,6 +246,39 @@ def select(self, field_paths: Iterable[str]) -> "BaseQuery":
...
+            end_at=self._evaluate_param(end_at, self._end_at),
+            all_descendants=self._evaluate_param(
+                all_descendants, self._all_descendants
+            ),
+        )
+
+    def _evaluate_param(self, value, fallback_value):
+        """Helper which allows `None` to be passed into `copy` and be set on the
+        copy instead of being misinterpreted as an unpassed parameter."""
+        return value if value != _not_passed else fallback_value
```
chrisrossi, line 280: Oops, sorry, should be `value is not _not_passed`. (You're testing that it is that specific object, not an equivalent object.)

craiglabenz: I'm not certain where this would materially change outcomes, since `object() != object()`, but, I've changed it over to the `id()`-based check of `is not` to be safe.

chrisrossi: It's more about saying what you mean, than the code working or not. A linter, for example, would give a warning if you said `value != None` and this is the same scenario.

chrisrossi (approved):

PR #407, opened 2021-07-29
feat: add support for recursive queries

craiglabenz:
Adds the ability for queries to begin fetching all nested documents _regardless of their collection name_. Previously, the only similar mechanic available was the `all_descendants` parameter (available via the `collection_group` class), which is imperfectly named and in fact only fetches all descendants _with the same `parent` name_. So, for example:

```
/users/alfred
/users/alfred/pets/scruffy
/users/alfred/users/alfreds-butler
```

In this case, running `db.collection_group('users').get()` would fetch `/users/alfred` and `/users/alfred/users/alfreds-butler`, but NOT `/users/alfred/pets/scruffy`.

Not that this is decidedly _not_ all of Alfred's descendants.

---

And thus: Recursive Queries ™️. Starting now, to fetch all three documents from the above hypothetical, a developer would run one of these two options:

```py
db.collection_group('users').recursive().get()
# or this similar statement with an identical outcome:
db.collections('users').recursive().get()
```

### DIFF
--- a/google/cloud/firestore_v1/_helpers.py
+++ b/google/cloud/firestore_v1/_helpers.py
@@ -76,6 +76,7 @@ class EmptyParam:
     
     When seen, the pseudo-singleton `empty_param` value below serves as a sentinel
     to indicate that the parameter was ignored by calling code."""
+
     def __eq__(self, other):
         return isinstance(other, EmptyParam)
 
--- a/google/cloud/firestore_v1/_helpers.py
+++ b/google/cloud/firestore_v1/_helpers.py
@@ -70,20 +70,6 @@ _GRPC_ERROR_MAPPING = {
 }
 
 
-class EmptyParam:
-    """Default value for optional parameters used to disambiguate between
-    calling code passing `None` versus ignoring the parameter entirely.
-    
-    When seen, the pseudo-singleton `empty_param` value below serves as a sentinel
-    to indicate that the parameter was ignored by calling code."""
-
-    def __eq__(self, other):
-        return isinstance(other, EmptyParam)
-
-
-empty_param = EmptyParam()
-
-
 class GeoPoint(object):
     """Simple container for a geo point value.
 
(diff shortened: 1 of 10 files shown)


google/cloud/firestore_v1/_helpers.py
```diff
@@ -139,14 +139,20 @@ def verify_path(path, is_collection) -> None:
...
         if num_elements % 2 == 0:
-            raise ValueError("A collection must have an odd number of path elements")
+            raise ValueError(
+                f"A collection must have an odd number of path elements. Encountered {path}"
+            )
     else:
         if num_elements % 2 == 1:
-            raise ValueError("A document must have an even number of path elements")
+            raise ValueError(
+                f"A document must have an even number of path elements. Encountered {path}"
```
crwilcox, line 154: Is this left over from testing? Do we think that customers will find this more helpful. there is a nonzero chance folks are relying on these messages is all, but I wouldn't recommend doing that anyways so I could be convinced to add this I think?

google/cloud/firestore_v1/base_collection.py
```diff
@@ -124,7 +124,10 @@ def document(self, document_id: str = None) -> DocumentReference:
         if document_id is None:
             document_id = _auto_id()
 
-        child_path = self._path + (document_id,)
+        # Append `self._path` and the passed document's  Id as long as the first
```
crwilcox, line 127:
```suggestion
        # Append `self._path` and the passed document's ID as long as the first
```

nit: extra space, Id -> ID

google/cloud/firestore_v1/_helpers.py
craiglabenz: Interesting thought. They are left over from development, but I left them in for the improved clarity. That said, I don't want to take on resolving exception message checking conflicts with this, so I'll revert.

tseaver: I don't think specific message texts are part of the API:  customers who are relying on them could get broken,  e.g., by minor Python version changes as well.

crwilcox (approved):

PR #413, opened 2021-08-03
feat: add retry support for BulkWriter errors

craiglabenz:
Contains the extra retry logic for BulkWriter errors.

The high level changes of this PR vs the previous state of `bulk-writer-release` are:

1. A new wrapping class, `BulkWriteOperation`, is introduced. This wraps all the arguments for a given BulkWriter operation (`create`, `update`, `set`, or `delete`), along with meta information like how many attempts have been made so far.
2. The in-progress `BulkWriteBatch` is now replaced with a simple list of `BulkWriteOperation` instances.
3. Handling around `on_write`, `on_error`, and `on_batch` callbacks have been improved.
4. All futures from the executor are tracked, allowing `flush()` to wait on their completion. This is important because some of the final requests could fail, leading to last-second retries that `flush()` must also detect and wait on.

### DIFF
--- a/google/cloud/firestore_v1/bulk_writer.py
+++ b/google/cloud/firestore_v1/bulk_writer.py
@@ -15,32 +15,59 @@
 """Helpers for efficiently writing large amounts of data to the Google Cloud
 Firestore API."""
 
+import bisect
 import collections
 import concurrent.futures
+import datetime
 import enum
 import functools
 import logging
 import time
-from typing import Callable, Dict, List, NoReturn, Optional, Union
+
+from typing import Callable, Dict, List, Optional, Union, TYPE_CHECKING
+
+from google.rpc import status_pb2  # type: ignore
 
 from google.cloud.firestore_v1 import _helpers
 from google.cloud.firestore_v1.base_document import BaseDocumentReference
-from google.cloud.firestore_v1.base_client import BaseClient
 from google.cloud.firestore_v1.bulk_batch import BulkWriteBatch
 from google.cloud.firestore_v1.rate_limiter import RateLimiter
 from google.cloud.firestore_v1.types.firestore import BatchWriteResponse
 from google.cloud.firestore_v1.types.write import WriteResult
 
+if TYPE_CHECKING:
+    from google.cloud.firestore_v1.base_client import BaseClient  # pragma: NO COVER
+
 
 logger = logging.getLogger(__name__)
 
 
-class SendMode(enum.Enum):
-    """Control flag for whether a BulkWriter should commit batches in the main
-    thread or hand that work off to an executor.
-    """
+class BulkRetry(enum.Enum):
+    """Indicator for what retry strategy the BulkWriter should use."""
 
+    # Common exponential backoff algorithm. This strategy is largely incompatible
+    # with the default retry limit of 15, so use with caution.
+    exponential = enum.auto()
+
+    # Default strategy that adds 1 second of delay per retry.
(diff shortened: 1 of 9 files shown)


tseaver:
@googlebot I consent.

google/cloud/firestore_v1/bulk_writer.py
```diff
@@ -155,15 +226,15 @@ class BulkWriter(AsyncBulkWriterMixin):
         db = firestore.Client()
         bulk_writer = db.bulk_writer()
 
-        # Attach an optional success listener. This will be called once per
-        # document.
+        # Attach an optional success listener to be called once per document.
         bulk_writer.on_write_result(
-            lambda reference, result: print(f'Saved {reference._document_path}')
+            lambda reference, result, bw: print(f'Saved {reference._document_path}')
```
crwilcox, line 231: just verifying this is intentional?

google/cloud/firestore_v1/bulk_writer.py
```diff
@@ -181,25 +252,51 @@ class BulkWriter(AsyncBulkWriterMixin):
...
+        client: Optional["BaseClient"] = None,
         options: Optional["BulkWriterOptions"] = None,
     ):
-        self._client = client
+        # Because `BulkWriter` instances are all synchronous/blocking on the
+        # main thread, instead using other threads for asynchrony, the
+        # `AsyncClient`'s various methods that return Futures are incompatible.
+        # `BulkWriter` parallelizes all of its network I/O without the developer
+        # having to worry about awaiting async methods, so we must convert an
+        # AsynClient into a plain Client instance.
```
crwilcox, line 263:
```suggestion
        # AsyncClient into a plain Client instance.
```

google/cloud/firestore_v1/bulk_writer.py
```diff
@@ -181,25 +252,51 @@ class BulkWriter(AsyncBulkWriterMixin):
...
-        self._is_sending: bool = False
 
-        self._success_callback: Optional[
-            Callable[[BaseDocumentReference, WriteResult, "BulkWriter"], NoReturn]
-        ] = None
-        self._batch_callback: Optional[
-            Callable[[BulkWriteBatch, BatchWriteResponse, "BulkWriter"], NoReturn]
-        ] = None
-        self._error_callback: Optional[Callable] = None
+        # Stores all currently pending futures for the purposes of the `flush`
```
crwilcox, line 287: This comment seems a bit odd to be attached to an empty list. Is there something else going on

google/cloud/firestore_v1/bulk_writer.py
```diff
@@ -213,19 +310,37 @@ def __init__(
 
         self._ensure_executor()
 
-    def _reset_batch(self) -> BulkWriteBatch:
-        self._batch = BulkWriteBatch(self._client)
-        return self._batch
+    @staticmethod
+    def _default_on_batch(
+        batch: BulkWriteBatch, response: BatchWriteResponse, bw: "BulkWriter",
```
crwilcox, line 315: `bw` doesn't seem like the pythonic name, though it isn't awful. It seems to be the name of a writer?

google/cloud/firestore_v1/bulk_writer.py
craiglabenz: Yep, that new parameter is real; though I've renamed it to `bulk_writer`.

crwilcox (approved):

PR #91, opened 2020-07-10
feat!: Begin using new microgenerator for v2 firestore

crwilcox:
Commits early on split off to separate out generated bits from edits to (hopefully) ease reviewing

PR #417, opened 2021-08-09
docs: fixed broken links to devsite

### LINKED ISSUE #328: Broken links in documentation
On this page:

https://googleapis.dev/python/firestore/latest/transforms.html

all the links beginning with https://cloud.google.com/firestore/docs/reference/ return a 404.

Also, each of the googleapis.dev pages has a "Report issue" link that leads here:

https://issuetracker.google.com/savedsearches/5337669

But when I reported the above issue there, I was [told it was the wrong place to report it](https://issuetracker.google.com/issues/181793700). So probably the "Report issue" link should be changed.

In general the reference documentation for the Python Firestore library has always been hard to use. In fact all the Firebase admin client libraries for different languages seem to have their reference documentation in completely different formats on different sites. The reference for the JavaScript web client library is much better. It would be great if all Firebase reference documentation worked like that.

craiglabenz:
Resolves #328

### DIFF
--- a/google/cloud/firestore_v1/transforms.py
+++ b/google/cloud/firestore_v1/transforms.py
@@ -72,7 +72,7 @@ class ArrayUnion(_ValueList):
     """Field transform: appends missing values to an array field.
 
     See:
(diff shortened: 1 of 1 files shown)


parthea (approved):

tseaver (commented):
Funny thing -- I actually had those links right before the microgen-switch in #91.

PR #420, opened 2021-08-11
feat: add recursive delete

craiglabenz:
Adds `recursive_delete()` method which deletes and document or collection and all sub-collections.

### DIFF
--- a/google/cloud/firestore_v1/async_client.py
+++ b/google/cloud/firestore_v1/async_client.py
@@ -43,13 +43,17 @@ from google.cloud.firestore_v1.async_document import (
     DocumentSnapshot,
 )
 from google.cloud.firestore_v1.async_transaction import AsyncTransaction
+from google.cloud.firestore_v1.field_path import FieldPath
 from google.cloud.firestore_v1.services.firestore import (
     async_client as firestore_client,
 )
 from google.cloud.firestore_v1.services.firestore.transports import (
     grpc_asyncio as firestore_grpc_transport,
 )
-from typing import Any, AsyncGenerator, Iterable, List
+from typing import Any, AsyncGenerator, Iterable, List, Optional, Union, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from google.cloud.firestore_v1.bulk_writer import BulkWriter  # pragma: NO COVER
 
 
 class AsyncClient(BaseClient):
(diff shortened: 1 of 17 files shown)


google/cloud/firestore_v1/async_query.py
```diff
@@ -126,6 +127,47 @@ def __init__(
             recursive=recursive,
         )
 
+    async def chunkify(
+        self, chunk_size: int
+    ) -> AsyncGenerator[List[DocumentSnapshot], None]:
+        # Catch the edge case where a developer writes the following:
+        # `my_query.limit(500).chunkify(1000)`, which ultimately nullifies any
```
crwilcox, line 134: This comment seems to be the opposite of the check `chunk_size < self._limit` (chunk size is larger in sample than limit)

google/cloud/firestore_v1/base_document.py
```diff
@@ -315,10 +315,10 @@ def _prep_collections(
 
     def collections(
         self, page_size: int = None, retry: retries.Retry = None, timeout: float = None,
-    ) -> NoReturn:
```
crwilcox, line 318: Is there a difference in these returns? Seems unrelated to the change otherwise?

google/cloud/firestore_v1/async_client.py
```diff
@@ -300,6 +304,84 @@ async def collections(
...
+    ) -> int:
+        """Recursion helper for `recursive_delete."""
+        from google.cloud.firestore_v1.bulk_writer import BulkWriter
+
+        bulk_writer = bulk_writer or BulkWriter()
+
+        num_deleted: int = 0
+
+        if isinstance(reference, AsyncCollectionReference):
+            chunk: List[DocumentSnapshot]
```
crwilcox, line 354: Is this necessary? It seems chunkify knows what type it returns, so it seems odd to declare it here as well?

google/cloud/firestore_v1/client.py
```diff
@@ -286,6 +291,87 @@ def collections(
...
+    ) -> int:
+        """Recursion helper for `recursive_delete."""
+        from google.cloud.firestore_v1.bulk_writer import BulkWriter
+
+        bulk_writer = bulk_writer or BulkWriter()
+
+        num_deleted: int = 0
+
+        if isinstance(reference, CollectionReference):
+            chunk: List[DocumentSnapshot]
```
crwilcox, line 342: Curious if you have thoughts to de-dupe this with the async version. It isn't *quite* copied, but it is irritatingly similar. Not something we have to do here, but would be interested to get your ideas here if you have any.

For instance, while the user interface is synchronous, could the library use the async version in this method for a speed improvement?

tseaver:
Rebasing against current `master` would clarify this PR, as well as likely appeasing the CLA bot (since the mixed commits are from the merged bulk writer PR).

google/cloud/firestore_v1/async_query.py
craiglabenz: Oh my goodness. I reversed this today in a moment of what I now see was confusion. Never was good at left and right, either.

google/cloud/firestore_v1/base_document.py
craiglabenz: There is! `NoReturn` tells IDEs that the function _never returns_, so any code appearing after such a method call is deemed unreachable. To that end, I cleaned up any `NoReturn` type hints I found on methods that my code was calling.

google/cloud/firestore_v1/async_client.py
craiglabenz: I'm not sure how much of the Python community uses this line, but (as you can see), I have been a bit recently for further clarity. I could go either way on it, though, if others have strong feelings.

google/cloud/firestore_v1/client.py
craiglabenz: Strong agree on the similarity being irritating. Let's chat.

google/cloud/firestore_v1/async_collection.py
```diff
@@ -72,6 +72,10 @@ def _query(self) -> async_query.AsyncQuery:
         """
         return async_query.AsyncQuery(self)
 
+    async def chunkify(self, chunk_size: int):
```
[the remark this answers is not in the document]
craiglabenz: It certainly could be. It seems like a helpful utility, but I would also be fine making it internal.

google/cloud/firestore_v1/async_query.py
craiglabenz: I even had the tests flipped 😭

google/cloud/firestore_v1/base_document.py
craiglabenz: One interesting detail is that you may be referencing is that this parent class implementation does nothing, so this method does actually _never return_. That said, other classes that can accept either `Document` or `AsyncDocument` instances might specify their type hints with `BaseDocument`, and in those methods, IDEs think any code that appears after calling `some_document.collections()` is unreachable.

crwilcox: I was sincerely curious, not secretly knowing the answer here :) Thanks for explaining!

craiglabenz: I only learned this myself working on BulkWriter when I finally got fed up with VS Code "getting confused" about certain code paths being unreachable. Like always, it was just PEBKAC.

craiglabenz:
As per offline conversations, I think this is ready to go, @crwilcox.

google/cloud/firestore_v1/async_client.py
```diff
@@ -300,6 +304,84 @@ async def collections(
...
+                [FieldPath.document_id()]
+            ).chunkify(chunk_size):
+                doc_snap: DocumentSnapshot
+                for doc_snap in chunk:
+                    num_deleted += 1
+                    bulk_writer.delete(doc_snap.reference)
+
+        elif isinstance(reference, AsyncDocumentReference):
+            col_ref: AsyncCollectionReference
+            async for col_ref in reference.collections():
```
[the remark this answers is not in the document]
crwilcox: Ah, it added that diff on the link.

```

func (c *Client) RecursiveDocumentQuery(documentPath string) Query {
	// make a query with parent set to this document, then collection id unset.
	docRef := c.Doc(documentPath)
	q := Query{c: c}.Select(DocumentID)

	// Collection id should be unset.
	q.collectionID = ""

	// Parent path should be configured to this document
	q.parentPath = docRef.Path
	q.path = q.parentPath

	q.allDescendants = true

	return q
}

```

google/cloud/firestore_v1/async_collection.py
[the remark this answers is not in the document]
crwilcox: If we have an example for how users would use it we could expose it, but generally I think we prefer to keep things internal if we aren't sure since it is a commitment to future design.

google/cloud/firestore_v1/client.py
crwilcox: https://github.com/googleapis/python-firestore/issues/425

crwilcox (approved):

PR #422, opened 2021-08-12
fix: remove unused requirement pytz

busunkim96:
I noticed firestore doesn't seem to use pytz for anything other than a unit test.


crwilcox (approved):

crwilcox:
Thanks for spotting this @busunkim96 !

PR #456, opened 2021-09-23
fix: harden 'query.stream' against retriable exceptions

### LINKED ISSUE #223: Port query retry from Node, Java to Python
To harden queries, we can restart a stream with the last received document as the cursor

Node Implementation: https://github.com/googleapis/nodejs-firestore/pull/1116
Java Implementation: https://github.com/googleapis/java-firestore/pull/426

tseaver:
Closes #223.

### DIFF
--- a/google/cloud/firestore_v1/query.py
+++ b/google/cloud/firestore_v1/query.py
@@ -20,6 +20,7 @@ a more common way to create a query than direct usage of the constructor.
 """
 from google.cloud import firestore_v1
 from google.cloud.firestore_v1.base_document import DocumentSnapshot
+from google.api_core import exceptions  # type: ignore
 from google.api_core import gapic_v1  # type: ignore
 from google.api_core import retry as retries  # type: ignore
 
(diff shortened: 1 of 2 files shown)


google/cloud/firestore_v1/query.py
```diff
@@ -244,15 +268,27 @@ def stream(
...
         )
 
-        response_iterator = self._client._firestore_api.run_query(
-            request=request, metadata=self._client._rpc_metadata, **kwargs,
-        )
+        last_snapshot = None
+
+        while True:
+            try:
+                response = next(response_iterator)
```
crwilcox, line 279: Could we inject this retry into `__iter__` somehow? this seems more generally useful to me.

tseaver: Because the underlying API call is a `unary_stream`, the `response_iterator` is an instance of `google.api_core.grpc_helpers._StreamingResponseIterator`: I guess we could modify its `__next__` to do something like this (it is already tweaked to cache the first result, for instance).

tseaver: I don't see how we get the applicable `retry` object inside there -- the `_wrapped` attribute isn't our `_GapicCallable`, but something altogether different.

crwilcox (approved):

PR #488, opened 2021-11-03
fix: recursive_delete index error on empty collection

noxxious:
Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-firestore/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

Fixes #487  🦕

PR #489, opened 2021-11-03
fix: handle empty last chunk correctly in 'Query._chunkify'

### LINKED ISSUE #487: Recursive delete fails with no documents
#### Environment details

  - OS type and version: Linux Manjaro, Kernel 5.15 (applies to any)
  - Python version: Python 3.9.6
  - pip version: 20.3.4
  - `google-cloud-firestore` version: 2.3.4

#### Steps to reproduce

  1. have no documents in the collection
  2. call Client.recursive_delete

#### Code example

```python
# example
```

#### Stack trace
```
Traceback (most recent call last):
  File "store.py", line 141, in drop_all
    self.client.recursive_delete( )
  File "google/cloud/firestore_v1/client.py", line 322, in recursive_delete
    return self._recursive_delete(
  File "google/cloud/firestore_v1/client.py", line 343, in _recursive_delete
    for chunk in (
  File "google/cloud/firestore_v1/query.py", line 198, in _chunkify
    last_document = snapshots[-1]
IndexError: list index out of range
```

Adding a simple check on the failing line for an empty array does the trick:
```python
last_document = snapshots[-1] if snapshots else None
```

tseaver:
Closes #487.
Supersedes #488.


noxxious:
In the #488 I also had a fix for the None case of the bulk_writter being not correctly initialized.

tseaver:
@noxxious Your change (calling `self.bulk_writer` rather than instantiating `BulkWriter()` directly) didn't change the behavior:  the client's `bulk_writer` method takes an optional `options` argument, but it isn't being passed.

noxxious:
Sure, passing options would have changed the client API interface. With the current API of bulk_writer being optional and current coded behavior of constructing bulk_writer, using a default constructor results in non-functional BulkWriter not bound to a client, hence all the recursive delete actions are not performed.

tseaver:
@noxxious Thanks for the follow-up!

It is indeed a problem that we have code constructing `BulkWriter()` without passing a client:  that argument should *not* be optional (now fixed in db9599b).

a2e5469 refactors the bits which construct the bulk writer if none is passed to the public`recursive_delete` methods:  the private `_recursive_delete` helpers now have it as a required argument, and the public methods construct it using `self.bulk_writer()`, as in your patch.

noxxious:
Glad to help.

chrisrossi (approved):
LGTM!

PR #496, opened 2021-11-18
feat: allow 'Collection.where("__name__", "in", ["hello", "world"])'

### LINKED ISSUE #421: Convert strings passed for an in query on a DocumentID to DocumentRefs for ease of use
When querying a collection with an in filter on a documentId, passing an array of strings will not work, instead returning an error

Repro:
```

client = firestore.Client("your-project-id")
collection = client.collection('0')

query = collection.where("__name__", "in", ["hello", "world"])
[i for i in query.stream()]
```

Node.js handles this today by validating the reference and as part of it doing a conversion. It would be convenient for go users if a similar conversion were done.
Referenced Node Code: https://github.com/googleapis/nodejs-firestore/blob/9a1a9dcd56dc125159be3ce7356a6ebce04519a9/dev/src/reference.ts#L1646

This code works today, so this is a matter of detecting the type mismatch with DocumentID queries and adjusting it on the user's behalf.
```

client = firestore.Client("crwilcox-firestore")
collection = client.collection('0')

docRefs = [
    collection.document("hello"),
    collection.document("world")
]
query = collection.where("__name__", "in", docRefs)
[i for i in query.stream()]
```

Related Go Issue: https://github.com/googleapis/google-cloud-go/issues/4151

tseaver:
Closes #421

### DIFF
--- a/google/cloud/firestore_v1/base_collection.py
+++ b/google/cloud/firestore_v1/base_collection.py
@@ -236,16 +236,29 @@ class BaseCollectionReference(object):
             field_path (str): A field path (``.``-delimited list of
                 field names) for the field to filter on.
             op_string (str): A comparison operation in the form of a string.
-                Acceptable values are ``<``, ``<=``, ``==``, ``>=``
-                and ``>``.
+                Acceptable values are ``<``, ``<=``, ``==``, ``>=``, ``>``,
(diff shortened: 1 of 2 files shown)


tseaver:
Superseded by #501.

PR #497, opened 2021-11-19
fix: stop / start stream after filter mismatch

### LINKED ISSUE #367: Collection on_snapshot incorrectly reports everything removed on internet reconnect
#### Environment details

  - OS type and version: Windows 10 21H1 build 19043.985
  - Python version: 3.9.4
  - pip version: 21.0.1
  - `google-cloud-firestore` version: 2.1.1

#### Steps to reproduce

  1. Setup a collection snapshot listener while connected to internet
  2. Disconnect from internet, wait a few seconds
  3. On a different device, remove a document from the collection
  4. Reconnect to internet, wait a few seconds
  5. Snapshot listener gets called indicating that **every** item has been removed, not just whatever has actually been removed

#### Code example

```python

def on_snapshot(snapshot, changes, read_time):
    print(f"Snapshot: [{', '.join(s.id for s in snapshot)}]")
    for change in changes:
        print(f"    {change.type.name}: {change.document.id}")

db = firestore.Client()
collection = db.collection("testCollection")
collection.on_snapshot(on_snapshot)
while True:
    pass
```

#### Other Info
I thought I had seen the issue occur even when skipping step 3 above, but have been unable to replicate that scenario.

My situation requires a degree of "offline support", which is when this came up. Individual document listeners don't seem to have any issues, but I'm expecting hundreds or thousands of documents and setting up a listener on each one seems overkill (and well beyond the recommended 100 max listeners outlined at [https://firebase.google.com/docs/firestore/best-practices#realtime_updates](https://firebase.google.com/docs/firestore/best-practices#realtime_updates)).

tseaver:
Closes #367.

### DIFF
--- a/google/cloud/firestore_v1/collection.py
+++ b/google/cloud/firestore_v1/collection.py
@@ -237,9 +237,5 @@ class CollectionReference(BaseCollectionReference):
             # Terminate this watch
             collection_watch.unsubscribe()
         """
-        return Watch.for_query(
-            self._query(),
-            callback,
-            document.DocumentSnapshot,
-            document.DocumentReference,
-        )
+        query = self._query()
+        return Watch.for_query(query, callback, document.DocumentSnapshot)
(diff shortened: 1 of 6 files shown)


tseaver:
Superseded by #502.

PR #500, opened 2021-12-07
refactor: use 'WhichOneof' for dispatch in 'Watch.on_snapshot'

tseaver:
Improve both performance and readability of `Watch.on_snaphot` by dispatching based on the `response_type` oneof:  requires using the "raw" protobuf messages, which we also use where feasible for speed.

Replace oddball testing-only `BackgroundConsumer` and `ResumableBidiRpc` arguments to `Watch` with "standard" mocking.

Use "real" protobuf messages for testing, versus oddball manual mocks.

~~Based on branch for PR #499 -- I will rebase after that PR merges.~~

### DIFF
--- a/google/cloud/firestore_v1/watch.py
+++ b/google/cloud/firestore_v1/watch.py
@@ -12,22 +12,24 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import logging
 import collections
-import threading
 from enum import Enum
 import functools
+import logging
+import threading
 
 from google.api_core.bidi import ResumableBidiRpc
 from google.api_core.bidi import BackgroundConsumer
-from google.cloud.firestore_v1.types import firestore
-from google.cloud.firestore_v1 import _helpers
-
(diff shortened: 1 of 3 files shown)


google/cloud/firestore_v1/watch.py
```diff
@@ -325,182 +372,122 @@ def _on_rpc_done(self, future):
...
+        if document_name.startswith(self._documents_pfx):
+            document_name = document_name[self._documents_pfx_len :]
+        return document_name
+
     def on_snapshot(self, proto):
-        """
-        Called everytime there is a response from listen. Collect changes
-        and 'push' the changes in a batch to the customer when we receive
-        'current' from the listen response.
+        """Process a response from the BIDI stream.
```
tswast, line 431: Let's avoid BIDI.

```suggestion
        """Process a response from the bi-directional gRPC stream.
```

tswast (approved):
Much cleaner! "oneof" for the win. Just a minor docstring suggestion

PR #509, opened 2021-12-27
feat: add 'DocumentSnapshot.__repr__'

gieddy660:
…pshot

### DIFF
--- a/google/cloud/firestore_v1/base_document.py
+++ b/google/cloud/firestore_v1/base_document.py
@@ -370,6 +370,9 @@ class DocumentSnapshot(object):
     def __hash__(self):
         return hash(self._reference) + hash(self.update_time)
 
+    def __repr__(self):
+        return repr(self.to_dict())
+
     @property
     def _client(self):
         """The client that owns the document reference for this snapshot.


tseaver (commented):
Thanks very much for the patch!

I find myself wondering whether the `__repr__` should include the snapshot's *metadata* (reference path, existence, create / read / update times) in addition to its *data*.

We likely also need one or more tests for the implementation.

gieddy660:
Regarding the metadata, while useful information, I feel that showing it all might create too much of a mess (the two DatetimeWithNanoseconds alone take more than 150 characters).
I also agree taht some tests might be needed, but being somewhat inexpirienced I wouldn't really know what to actually test (after all I only added 2 lines, and DocumentSnapshot.to_dict() isn't all that complicated either).

tseaver:
@gieddy660 
> Regarding the metadata, while useful information, I feel that showing it all might create too much of a mess (the two DatetimeWithNanoseconds alone take more than 150 characters).

Given that the *data* can be arbitrarily huge, I'm not sure how to judge.  @tswast do you have an opinion here?

tswast:
> Given that the data can be arbitrarily huge, I'm not sure how to judge. @tswast do you have an opinion here?

I'm wary of including data in repr for privacy reasons as well. `repr` values are likely to end up in exceptions and logs.

With only minimal knowledge of Firestore, it seems to me that `reference` and `read_time` are actually the key / unique data importance to DocumentSnapshot and pose minimal privacy risk.

PR #521, opened 2022-01-20
chore: remove googleapis/firestore-dpe from codeowners

parthea:
The correct team is already in codeowners as googleapis/api-firestore


kolea2 (approved):

parthea:
The system test failures are being tracked in https://github.com/googleapis/python-firestore/issues/518

PR #604, opened 2022-06-07
fix(deps): require protobuf<4.0.0 on v1 branch

parthea:


dandhlee (approved):
Kokoro failing for `unit-3.5` and couldn't find Kokoro config file for system test 3.7.

PR #624, opened 2022-08-11
fix(deps): allow protobuf < 5.0.0

parthea:
fix(deps): require proto-plus >= 1.22.0


unforced (approved):
Looks like a foreign key error that needs resolved before this can be submitted.

PR #673, opened 2023-01-04
feat: Support the Count aggregation query

Mariatta:


danieljbruce (commented):
First pass.

tests/system/test_system.py
```diff
@@ -536,6 +536,13 @@ def query_docs(client):
         operation()
 
 
+@pytest.fixture
```
kolea2, line 539: looks good. can we add some tests running this with a transaction?

kolea2 (dismissed):

tests/system/test_system.py
```diff
@@ -1617,3 +1624,130 @@ def test_repro_391(client, cleanup):
...
+
+
+def test_count_query_get_multiple_aggregations(query):
+    count_query = query.count(alias="total").count(alias="all")
+
+    result = count_query.get()
+    assert len(result[0]) == 2
+
+    for r in result[0]:
+        assert r.alias in ["total", "all"]
```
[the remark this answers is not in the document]
Mariatta: From my tests, the order of the result has been inconsistent. It doesn't always return it in the order of "total" and then "all".

I've now adjusted this test to assert that:
- there are two elements returned in the result
- the returned result has two values and they are both unique
- the result contains both aliases: "total" and "all"

tests/system/test_system.py
Mariatta: I've added a new test that uses transaction in both test_system and test_system_async.

google/cloud/firestore_v1/async_aggregation.py
```diff
@@ -0,0 +1,122 @@
...
+        return result
+
+    async def stream(
+        self,
+        transaction=None,
+        retry: Union[
+            retries.Retry, None, gapic_v1.method._MethodDefault
+        ] = gapic_v1.method.DEFAULT,
+        timeout: float | None = None,
+    ):  # -> Generator[document.DocumentSnapshot, Any, None]: AsyncGenerator
```
kolea2, line 84: should this be commented out?

google/cloud/firestore_v1/base_aggregation.py
```diff
@@ -0,0 +1,199 @@
+from __future__ import annotations
```
kolea2, line 1: missing header

kolea2 (commented):
lgtm to me after header fix and PTAL at last comment.

Would like someone from yoshi-python to review as I've only looked from an API perspective. Thanks!

google/cloud/firestore_v1/base_aggregation.py
Mariatta: Added.

google/cloud/firestore_v1/async_aggregation.py
Mariatta: It actually should be uncommented so we have type annotations. I have fixed this.

google/cloud/firestore_v1/aggregation.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def _get_stream_iterator(self, transaction, retry, timeout):
+        """Helper method for :meth:`stream`."""
+        request, kwargs = self._prep_stream(
+            transaction,
+            retry,
+            timeout,
+        )
+
+        response_iterator = self._client._firestore_api.run_aggregation_query(
```
nicain, line 85: Nit: return directly here, no need for the temporary variable `response_iterator`

Mariatta: Thanks, updated.

tests/system/test_system.py
```diff
@@ -46,11 +47,13 @@ def _get_credentials_and_project():
     if FIRESTORE_EMULATOR:
         credentials = EMULATOR_CREDS
         project = FIRESTORE_PROJECT
-    else:
+    elif FIRESTORE_CREDS:
         credentials = service_account.Credentials.from_service_account_file(
             FIRESTORE_CREDS
         )
         project = FIRESTORE_PROJECT or credentials.project_id
+    else:
```
kolea2, line 55: would like someone from @googleapis/yoshi-python to take a look at this specifically

kolea2 (approved):
lgtm after @googleapis/yoshi-python takes a look as well. Thanks!

tests/system/test_system.py
meredithslota: @kolea2 did anyone from @googleapis/yoshi-python review this? I can't tell from the conversation.

kolea2: I believe @nicain did!

PR #688, opened 2023-02-08
fix: type hinting in collection.py

### LINKED ISSUE #687: Pylance Typing for collections
Type hinting does not take into account default `None` values for most fields, which breaks Pylance linting

For example, a simple

```python
    db.collection(f"mycollection").stream())
```

yields an error from Pylance

```
Argument of type "None" cannot be assigned to parameter "transaction" of type "Transaction" in function "stream"
  Type "None" cannot be assigned to type "Transaction"Pylance[reportGeneralTypeIssues](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportGeneralTypeIssues)
```

I saw a similar error in https://github.com/googleapis/python-firestore/issues/570, I think most of the type hinting should be reviewed.

#### Environment details

  - OS type and version: MacOS 12.4
  - Python version: 3.11.1
  - pip version: pip 23.0
  - `google-cloud-firestore` version: 2.9.1

#### Steps to reproduce

In VSCode : 
```python
    db.collection(f"mycollection").stream())
```

Sugihiru:
Mostly adding `| None` on all fields that may have `None` as a default value.

Fixes #687 🦕

### DIFF
--- a/google/cloud/firestore_v1/collection.py
+++ b/google/cloud/firestore_v1/collection.py
@@ -79,9 +79,9 @@ class CollectionReference(BaseCollectionReference):
     def add(
         self,
         document_data: dict,
-        document_id: str = None,
+        document_id: str | None = None,
         retry: retries.Retry = gapic_v1.method.DEFAULT,
-        timeout: float = None,
+        timeout: float | None = None,
     ) -> Tuple[Any, Any]:
         """Create a document in the Firestore database with the provided data.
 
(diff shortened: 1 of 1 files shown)


Mariatta (approved):
Thank you.

Mariatta (changes requested):
The `|` syntax is only available in Python 3.10+. Since we still support Python 3.7+, then this should be using `Union` instead of `|`. 

e.g
```
from typing import Union

...
document_id: Union[str, None]
```

Mariatta (approved):
Thanks!

PR #698, opened 2023-03-20
feat: OR Query implementation

Mariatta:
Introduce new Filter classes:

FieldFilter
And
Or

Add `filter` keyword arg to `Query.where()`
The positional arguments in `Query.where()` are now optional. `UserWarning` is now emitted when using `where()` without keyword args.

### DIFF
--- a/google/cloud/firestore_v1/base_collection.py
+++ b/google/cloud/firestore_v1/base_collection.py
@@ -23,6 +23,7 @@ from google.cloud.firestore_v1.base_aggregation import BaseAggregationQuery
 
 
 from typing import (
+    Optional,
     Any,
     AsyncGenerator,
     Coroutine,
@@ -113,7 +114,7 @@ class BaseCollectionReference(object):
     def _aggregation_query(self) -> BaseAggregationQuery:
         raise NotImplementedError
 
-    def document(self, document_id: str = None) -> DocumentReference:
+    def document(self, document_id: Optional[str] = None) -> DocumentReference:
         """Create a sub-document underneath the current collection.
 
         Args:
@@ -160,9 +161,9 @@ class BaseCollectionReference(object):
     def _prep_add(
         self,
         document_data: dict,
-        document_id: str = None,
-        retry: retries.Retry = None,
-        timeout: float = None,
+        document_id: Optional[str] = None,
+        retry: Optional[retries.Retry] = None,
+        timeout: Optional[float] = None,
     ) -> Tuple[DocumentReference, dict]:
         """Shared setup for async / sync :method:`add`"""
         if document_id is None:
@@ -176,17 +177,17 @@ class BaseCollectionReference(object):
     def add(
         self,
         document_data: dict,
-        document_id: str = None,
-        retry: retries.Retry = None,
-        timeout: float = None,
+        document_id: Optional[str] = None,
+        retry: Optional[retries.Retry] = None,
+        timeout: Optional[float] = None,
     ) -> Union[Tuple[Any, Any], Coroutine[Any, Any, Tuple[Any, Any]]]:
         raise NotImplementedError
 
     def _prep_list_documents(
         self,
-        page_size: int = None,
-        retry: retries.Retry = None,
-        timeout: float = None,
+        page_size: Optional[int] = None,
+        retry: Optional[retries.Retry] = None,
+        timeout: Optional[float] = None,
     ) -> Tuple[dict, dict]:
         """Shared setup for async / sync :method:`list_documents`"""
         parent, _ = self._parent_info()
@@ -206,9 +207,9 @@ class BaseCollectionReference(object):
 
     def list_documents(
         self,
-        page_size: int = None,
-        retry: retries.Retry = None,
-        timeout: float = None,
+        page_size: Optional[int] = None,
+        retry: Optional[retries.Retry] = None,
+        timeout: Optional[float] = None,
     ) -> Union[
         Generator[DocumentReference, Any, Any], AsyncGenerator[DocumentReference, Any]
     ]:
@@ -236,7 +237,14 @@ class BaseCollectionReference(object):
         query = self._query()
         return query.select(field_paths)
 
-    def where(self, field_path: str, op_string: str, value) -> BaseQuery:
+    def where(
+        self,
+        field_path: Optional[str] = None,
+        op_string: Optional[str] = None,
+        value=None,
+        *,
+        filter=None
+    ) -> BaseQuery:
         """Create a "where" query with this collection as parent.
 
         See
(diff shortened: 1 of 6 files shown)


google/cloud/firestore_v1/base_collection.py
```diff
@@ -160,9 +161,9 @@ def _parent_info(self) -> Tuple[Any, str]:
     def _prep_add(
         self,
         document_data: dict,
-        document_id: str = None,
-        retry: retries.Retry = None,
-        timeout: float = None,
+        document_id: Optional[str] = None,
```
kolea2, line 164: why are all of these changing to optional?

tests/system/test_system.py
```diff
@@ -1818,8 +1850,71 @@ def test_count_query_in_transaction(client, cleanup):
...
+        if value > 5:
+            gt_5 += 1
+        if value < 10:
+            lt_10 += 1
+
+    assert gt_5 > 0
+    assert lt_10 > 0
+
+
+def test_query_with_complex_composite_filter(query_docs):
```
kolea2, line 1899: would be good to also add some transaction tests

kolea2 (changes requested):

google/cloud/firestore_v1/base_query.py
```diff
@@ -665,12 +802,24 @@ def _filters_pb(self) -> StructuredQuery.Filter:
         if num_filters == 0:
             return None
         elif num_filters == 1:
-            return _filter_pb(self._field_filters[0])
+            filter = self._field_filters[0]
+            if isinstance(filter, query.StructuredQuery.CompositeFilter):
+                return query.StructuredQuery.Filter(composite_filter=filter)
```
danieljbruce, line 807: This function returns a protobuf?

danieljbruce (commented):
Looks good. Just a few comments.

google/cloud/firestore_v1/base_collection.py
Mariatta: it is already "optional" before. By adding the default value (`document_id=None`), it means the `document_id` does not need to be passed when we call `_prep_add`. The `Optional[str]` syntax means: it can either be `str` type or `None` type.

So it was in fact a bug before that the annotation was only `str` and not `Optional[str]`.

google/cloud/firestore_v1/base_query.py
```diff
@@ -315,36 +422,66 @@ def where(self, field_path: str, op_string: str, value) -> "BaseQuery":
...
             )
-        elif isinstance(value, (transforms.Sentinel, transforms._ValueList)):
-            raise ValueError(_INVALID_WHERE_TRANSFORM)
-        else:
-            filter_pb = query.StructuredQuery.FieldFilter(
-                field=query.StructuredQuery.FieldReference(field_path=field_path),
-                op=_enum_from_op_string(op_string),
-                value=_helpers.encode_value(value),
+
+        field_path_module.split_field_path(field_path)  # raises
```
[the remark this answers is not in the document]
Mariatta: Hmm, this seems to have always been there since this code was added: https://github.com/googleapis/python-firestore/blame/main/google/cloud/firestore_v1/base_query.py#L246

Perhaps a leftover debugging info. I can remove it.

google/cloud/firestore_v1/base_query.py
```diff
@@ -315,36 +422,66 @@ def where(self, field_path: str, op_string: str, value) -> "BaseQuery":
...
+                filter_pb = query.StructuredQuery.FieldFilter(
+                    field=query.StructuredQuery.FieldReference(field_path=field_path),
+                    op=_enum_from_op_string(op_string),
+                    value=_helpers.encode_value(value),
+                )
 
-        new_filters = self._field_filters + (filter_pb,)
+            new_filters += (filter_pb,)
+        elif isinstance(filter, BaseFilter):
+            new_filters += (filter._to_pb(),)
```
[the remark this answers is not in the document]
Mariatta: `new_filters` is a local variable only on the `where` function. User will not interact with it directly.

google/cloud/firestore_v1/base_query.py
```diff
@@ -651,7 +788,7 @@ def end_at(
             document_fields_or_snapshot, before=False, start=False
         )
 
-    def _filters_pb(self) -> StructuredQuery.Filter:
+    def _filters_pb(self) -> Optional[StructuredQuery.Filter]:
```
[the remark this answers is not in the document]
Mariatta: The `Optional` is a type annotation in Python, and it's more of a documentation change than behavioral change.
The `Optional[sometype]` means that it can either be `sometype` or `None`.
This function was already returning `None` before (in line 666 below)
```
    if num_filters == 0:
        return None
```
So this is not a change in the return value. It was actually a bug that the return value was not typed as `Optional`. This change fixed the bug and make it accurate.

tests/system/test_system.py
Mariatta: I have added system tests (sync and async) using transaction

google/cloud/firestore_v1/base_collection.py
Mariatta: More accurately, it is not a bug, but a change in the `mypy` type checker itself.
Prior to `mypy` 0.991, implicit optional was allowed. 
```
def foo(s: str = None) -> int: 
```
In the above example, the `s` parameter is optional (can be None), but the type was specified as `str`. This is what's considered as "implicit" optional.

Starting mypy 0.991, by default `mypy` will only accept "explicit" optional type.
```
def foo(s: str = Optional[str] = None): ->:
```
Which explicitly states that `s` parameter must either be a string `str` or `None`.

tests/system/test_system.py
```diff
@@ -542,13 +544,13 @@ def query_docs(client):
 @pytest.fixture
 def query(query_docs):
     collection, stored, allowed_vals = query_docs
-    query = collection.where("a", "==", 1)
+    query = collection.where(filter=FieldFilter("a", "==", 1))
```
Mariatta, line 547: To the reviewers, this is an example of how the change in the `where` function.
It adds the keyword only arg `filter`, and you can see how it works before vs now.

tests/system/test_system.py
```diff
@@ -1814,12 +1846,116 @@ def test_count_query_in_transaction(client, cleanup):
...
 
-    query = collection.where("a", "==", 1)
+    query = collection.where(filter=FieldFilter("a", "==", 1))
     count_query = query.count()
     result = count_query.get()
     for r in result[0]:
         assert r.value == 2  # there are still only 2 docs
+
+
+def test_query_with_and_composite_filter(query_docs):
```
Mariatta, line 1860: The test cases below shows how to build composite filters using And and Or and how to pass them to the `filter` argument in `where`.

google/cloud/firestore_v1/base_collection.py
kolea2: awesome, thanks for the explanation!

tests/system/test_system.py
```diff
@@ -1814,12 +1846,116 @@ def test_count_query_in_transaction(client, cleanup):
...
+        if result.get("stats.sum") == 0:
+            sum_0 += 1
+        if result.get("stats.sum") == 4:
+            sum_4 += 1
+
+    assert sum_0 > 0
+    assert sum_4 > 0
+
+
+def test_or_query_in_transaction(client, cleanup):
```
kolea2, line 1920: looking for a few more nested test examples in all cases like: `a ==1 || (b > 5 && a == 2)`, `a == 1 && (b > 5 || b < 2)`

Mariatta: Added some more complex filtering above under `test_query_with_complex_composite_filter`.

I could not add more complex filter test using transaction because it would require adding index for each of the UNIQUE_RESOURCE_ID (different one each time I run the test, meaning new index for each time the test is run).

So I simply added running the test with complex And/or without transaction in `test_query_with_complex_composite_filter`.

I added the same test to the async test as well.

kolea2 (approved):

google/cloud/firestore_v1/base_query.py
```diff
@@ -315,36 +422,66 @@ def where(self, field_path: str, op_string: str, value) -> "BaseQuery":
...
+
+        if field_path is not None and op_string is not None:
+            if filter is not None:
+                raise ValueError(
+                    "Can't pass in both the positional arguments and 'filter' at the same time"
+                )
+            warnings.warn(
+                "Detected filter using positional arguments. Prefer using the 'filter' keyword argument instead.",
+                UserWarning,
+                stacklevel=2,
```
andrewsg, line 453: I'm not sure I understand stacklevel as it's used here - doesn't that mean that "where" will not be displayed as the origin of the warning?

tests/system/test_system.py
andrewsg: The old behavior without the filter arg is still expected to work, correct? Are there any system tests that still validate the old behavior or is that unit test only now?

google/cloud/firestore_v1/base_query.py
Mariatta: When I ran the code, with stacklevel 2, it shows the `where` as the source of issue.

Output using `stacklevel=2`

> /Users/mariattaw/google-github/python-firestore/venv/lib/python3.11/site-packages/google/cloud/firestore_v1/base_collection.py:290: UserWarning: Detected filter using positional arguments. Prefer using the 'filter' keyword argument instead.
  return query.where(field_path, op_string, value)


When I use stacklevel = 1, it shows the `warnings.warn` as the source of issue

Output of using stacklevel = 1
> /Users/mariattaw/google-github/python-firestore/venv/lib/python3.11/site-packages/google/cloud/firestore_v1/base_query.py:450: UserWarning: Detected filter using positional arguments. Prefer using the 'filter' keyword argument instead.
  warnings.warn(


I think it's clearer to show the warning with stacklevel 2 so user knows that `where` has a different signature now. What do you think?

andrewsg: Interesting, so the "where" is in the headline of the stack trace. That does make sense. I'm okay with either and defer to your judgement. Thanks for your response.

andrewsg (approved):

tests/system/test_system.py
Mariatta: I added additional system tests here: https://github.com/googleapis/python-firestore/pull/698/commits/eff93287c649a0798a94f86e35c78534314722e2

PR #715, opened 2023-05-10
feat: Sum/Avg aggregation queries

Mariatta:
Adds the ability to perform sum/avg aggregation query through:
- query.sum(),
- query.avg(),
- async_query.sum(),
- async_query.avg()


kolea2 (changes requested):

tests/system/test_system.py
```diff
@@ -1881,6 +1881,216 @@ def test_count_query_stream_empty_aggregation(query, database):
...
+    avg_query = query.avg("stats.product", alias="total").avg(
+        "stats.product", alias="all"
+    )
+
+    for result in avg_query.stream():
+        for aggregation_result in result:
+            assert aggregation_result.alias in ["total", "all"]
+
+
 @firestore.transactional
```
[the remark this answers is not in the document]
daniel-sanche: I added a couple, let me know what you think

tests/system/test_system.py
```diff
@@ -1881,72 +1881,214 @@ def test_count_query_stream_empty_aggregation(query, database):
     assert "Aggregations can not be empty" in exc_info.value.message
 
 
-@firestore.transactional
-def create_in_transaction(collection_id, transaction, cleanup):
-    collection = client.collection(collection_id)
+@pytest.mark.parametrize("database", [None, FIRESTORE_OTHER_DB], indirect=True)
```
kolea2, line 1884: the diff is a little confusing to view here. Were all the changes made for SUM/AVG? is it possible we can refactor some out into a separate PR?

kolea2 (commented):
lgtm with one question around tests

tests/system/test_system.py
daniel-sanche: I changed `test_count_query_in_transaction` to be a more general, parameterized `test_aggregation_query_in_transaction` test. That also lead to some clean-up in the other query-in-transaction tests, since there were some (duplicated) helpers that were now only used in one place, and it made more sense to move the logic into that test itself. And I wanted to keep the related tests fairly consistent as well

kolea2: ok cool, thanks for the explanation!

google/cloud/firestore_v1/aggregation.py
```diff
@@ -130,7 +130,6 @@ def stream(
             :class:`~google.cloud.firestore_v1.base_aggregation.AggregationResult`:
             The result of aggregations of this query
         """
-
```
kolea2, line 133: nit - can be reverted

kolea2 (dismissed):

kolea2 (approved):

google/cloud/firestore_v1/async_query.py
```diff
@@ -231,6 +231,40 @@ def count(
         """
         return AsyncAggregationQuery(self).count(alias=alias)
 
+    def sum(
```
kolea2, line 234: similar to https://github.com/googleapis/java-firestore/pull/1387/files#diff-fdd9700384056ce568a3fd41266eb1dfc5487c620bade78cde24dd677d7016b5R83, should we also have a constructor that accepts a `FieldPath`?

kolea2 (changes requested):

google/cloud/firestore_v1/base_collection.py
```diff
@@ -507,6 +507,30 @@ def count(self, alias=None):
         """
         return self._aggregation_query().count(alias=alias)
 
+    def sum(self, field_ref: str, alias=None):
```
kolea2, line 510: does FieldPath need to be appended here as well?

daniel-sanche: good catch! Fixed

(The type checking system should have caught this, but it looks like it's not running properly. I'll have to fix that in a separate PR)

kolea2: sgtm, please file an issue for tracking!

daniel-sanche: I have a bug for it here: https://github.com/googleapis/python-firestore/issues/773

daniel-sanche:
Note: the new `*_with_start_at` tests are expected to fail for now. PR should be blocked until they are ready

kolea2 (approved):

PR #716, opened 2023-05-11
docs: deprecate google.cloud.firestore_v1.rate_limiter.utcnow

Mariatta:
These will be deprecated starting in Python 3.12.

Replaced:
- `datetime.datetime.utcnow()` with `datetime.datetime.now(tz=datetime.timezone.utc)`

- `datetime.utcfromtimestamp()` with `datetime.fromtimestamp(tz=datetime.timezone.utc)`

Source: https://github.com/python/cpython/issues/103857

BEGIN_COMMIT_OVERRIDE
fix: Remove used of deprecated `datetime.datetime.utcnow()` and `datetime.utcfromtimestamp()`
docs: deprecate google.cloud.firestore_v1.rate_limiter.utcnow
END_COMMIT_OVERRIDE


google/cloud/firestore_v1/rate_limiter.py
```diff
@@ -17,7 +17,7 @@
 
 
 def utcnow():
-    return datetime.datetime.utcnow()
+    return datetime.datetime.now(tz=datetime.timezone.utc)
```
pganssle, line 20: Note that this changes a public interface. I'm guessing it won't break anyone if people are consistently doing things like:

```python
dt_end = rate_limiter.utcnow() + timedelta(seconds=30)
while rate_limiter.utcnow() < dt_end:
    sleep(1)
```

But if they are mixing between `rate_limiter.utcnow()` and `datetime.utcnow()` or some other thing that generates a naïve datetime, this will cause `TypeError`s.

(I'm also not familiar with this library's conventions, so maybe this isn't really a public interface anyway).

On the plus side, it seems like some of [the other places in the GCP framework](https://github.com/googleapis/python-cloud-core/blob/44de0d893daab18b74de9aa032f7e580a2e47d22/google/cloud/_helpers/__init__.py#L190) are treating naïve datetimes as UTC anyway, so that will probably reduce the blast radius here.

tests/unit/v1/test_rate_limiter.py
```diff
@@ -19,7 +19,7 @@
 
 # Pick a point in time as the center of our universe for this test run.
 # It is okay for this to update every time the tests are run.
-fake_now = datetime.datetime.utcnow()
+fake_now = datetime.datetime.now(tz=datetime.timezone.utc)
```
pganssle, line 22: This might be outside the scope of the PR you want to make here, but you may want to consider using [`time-machine`](https://github.com/adamchainz/time-machine) or, if that's infeasible for some reason, [`freezegun`](https://github.com/spulec/freezegun) instead of the current mocking approach.

The way it works now, you need to make sure that the thing you are patching returns something with the correct "awareness" (i.e. returns a naïve datetime or aware datetime), but with one of those frameworks, the awareness of the result is dictated by the `datetime` methods that you called, as it would be if you weren't mocking.

pganssle (approved):
Looks good from what I can tell, except for the change to the public interface, which you'd know more about than me.

tests/unit/v1/test_rate_limiter.py
Mariatta: Yeah I think this may be an improvement to be done as a separate PR

google/cloud/firestore_v1/rate_limiter.py
parthea, line 20:
```suggestion
    # We used datetime.utcnow() before, since it's deprecated from python 3.12,
    # we are using datetime.now(timezone.utc) now. "utcnow()" is offset-native
    # (no timezone info), but "now()" is offset-aware (with timezone info).
    # This will cause datetime comparison problem. For backward compatibility,
    # we need to remove the timezone info.
    now = datetime.datetime.now(datetime.timezone.utc)
    now = now.replace(tzinfo=None)
    return now
```

Recommending that we make the same change as `google-auth` for backward compatibility
https://github.com/googleapis/google-auth-library-python/pull/1413/files

parthea: My initial thought is that since the method name doesn't have a single leading underscore it could have been used. I think we should ensure backward compatibility. For example, it's `def utcnow():` and not `def _utcnow():`

https://peps.python.org/pep-0008/#method-names-and-instance-variables
`Use one leading underscore only for non-public methods`

pganssle: I was just going off of Mariatta's comment here, since I raised the same concern initially: https://github.com/googleapis/python-firestore/pull/716#discussion_r1191365451

It is arguable that this function is public, though also it suffers from the same problems that `datetime.utcnow` suffers from, so if you're going to have it continue to return a naïve datetime, it should probably have a deprecation warning anyway.

The linked commit definitely seems to be doing the wrong thing, since it's re-creating `datetime.utcnow` in a private function.

daniel-sanche: I don't have the full context on the issue, but this `utcnow` function is just a one-liner, and only used in a few places. Can we mark it as deprecated and just replace its internal usages with `datetime.datetime.now(datetime.timezone.utc)?`

I feel like keeping it around is more trouble than its worth

daniel-sanche (approved):
LGTM

google/cloud/firestore_v1/rate_limiter.py
parthea: Deprecated `google.cloud.firestore_v1.rate_limiter.utcnow` in 075e999a4638fd137e65757f8d5a65a3aa3fcef9

tests/unit/v1/test_rate_limiter.py
parthea: Switched to `freezegun` in 5f9f04f1eae89956cbcfe3c08e41638ac976e80f

google/cloud/firestore_v1/rate_limiter.py
parthea: Deprecated `google.cloud.firestore_v1.rate_limiter.utcnow` in 075e999a4638fd137e65757f8d5a65a3aa3fcef9

parthea:
I filed https://github.com/googleapis/python-firestore/issues/812 for the issue `Using the synchronous google.api_core.retry.Retry with asynchronous calls may lead to unexpected results. Please use google.api_core.retry_async.AsyncRetry instead.` which appears in the failing presubmit `Kokoro Prerelease Dependencies`

PR #717, opened 2023-05-11
feat: Multi db test parametrization

Mariatta:
The changes are:

- adding test parametrization wherever the client is used. It will run the test against the default db, and a named db named db can only be passed to the Client, and the underlying queries and aggregation queries will retrieve the database id from the Client.


tests/system/test__helpers.py
```diff
@@ -12,3 +12,5 @@
 UNIQUE_RESOURCE_ID = unique_resource_id("-")
 EMULATOR_CREDS = EmulatorCreds()
 FIRESTORE_EMULATOR = os.environ.get(_FIRESTORE_EMULATOR_HOST) is not None
+
+FIRESTORE_OTHER_DB = os.environ.get("SYSTEM_TESTS_DATABASE", "mw-second-db")
```
kolea2, line 16: this should be changed?

kolea2 (changes requested):
I dont think the ITs were run since this is a draft PR - can we convert to ready for review?

tests/system/test__helpers.py
vishwarajanand: Fixed it.

tests/unit/v1/test_client.py
```diff
@@ -36,26 +37,42 @@ def _make_credentials():
 
 def _make_default_client(*args, **kwargs):
```
kolea2, line 38: it would be great to add one additional test without parameterization testing client initialization with:
1. not setting the database
2. setting the database
3. setting the database to None

and asserting the database is set correctly.

vishwarajanand: @kolea2 there are 2 approaches to tackle this:

1. (current - neater approach but tests only client initialization) Adds `test_client_constructor_without_db` where DB isn't set and the database is asserted to equal `DEFAULT_DATABASE`

2. (Code become less readable, but tests are more exhaustive) we can add a check in `_make_default_client` to not use `database` param when it's an empty string during the client initialization. 

I choose the first approach for readability, but I have no qualms with the second one too. Leaving this comment open for now for Python reviewers.

kolea2 (approved):
lgtm, will defer to @googleapis/yoshi-python for the remaining review!

tests/system/test__helpers.py
```diff
@@ -12,3 +12,4 @@
 UNIQUE_RESOURCE_ID = unique_resource_id("-")
 EMULATOR_CREDS = EmulatorCreds()
 FIRESTORE_EMULATOR = os.environ.get(_FIRESTORE_EMULATOR_HOST) is not None
+FIRESTORE_OTHER_DB = os.environ.get("SYSTEM_TESTS_DATABASE", "system-tests-named-db")
```
vishwarajanand, line 15: I've created a DB into `abstract-disk-227418` project where the tests run.

```
curl -X POST https://firestore.googleapis.com/v1/projects/abstract-disk-227418/databases\?databaseId\=system-tests-named-db \
-H "Authorization: Bearer $(gcloud auth print-access-token)" \
-H "Content-type: application/json" \
--data '{"type":"FIRESTORE_NATIVE", "locationId":"nam5"}'
```

parthea (approved):

Issue #765, opened 2023-09-20
transactions that aren't retried don't rollback


daniel-sanche:
in [Transaction.__call__](https://github.com/googleapis/python-firestore/blob/9840d4345dca49ba610ae8e3ce4955678c18c49c/google/cloud/firestore_v1/transaction.py#L280), rollback is called [when retries are exhausted](https://github.com/googleapis/python-firestore/blob/9840d4345dca49ba610ae8e3ce4955678c18c49c/google/cloud/firestore_v1/transaction.py#L313), or [when pre-commit fails](https://github.com/googleapis/python-firestore/blob/9840d4345dca49ba610ae8e3ce4955678c18c49c/google/cloud/firestore_v1/transaction.py#L249), but other exceptions cause an [immediate raise](https://github.com/googleapis/python-firestore/blob/9840d4345dca49ba610ae8e3ce4955678c18c49c/google/cloud/firestore_v1/transaction.py#L278), without any rollback.

To fix this, the entire `__call__` function should be wrapped in a `try...except` block, that causes a rollback before any exceptions are raised out of the transaction

kolea2:
It looks like this was fixed in https://github.com/googleapis/python-firestore/pull/767, closing. @daniel-sanche please reopen if needed. Thanks!

PR #767, opened 2023-09-20
fix: ensure transactions rollback on failure

### LINKED ISSUE #766: actionable error: some errors related to rpc calls are missing debug info
Looking at the exceptions raised in python-firestore, I found a couple places where custom ValueErrors are raised after failed rpc calls, discarding all the useful server debugging information
- [Transactional.\__call\_\_](https://github.com/googleapis/python-firestore/blob/9840d4345dca49ba610ae8e3ce4955678c18c49c/google/cloud/firestore_v1/transaction.py#L315)
- [Transactional._pre_commit](https://github.com/googleapis/python-firestore/blob/9840d4345dca49ba610ae8e3ce4955678c18c49c/google/cloud/firestore_v1/transaction.py#L247)
- [AsyncTransactional.\__call\_\_](https://github.com/googleapis/python-firestore/blob/9840d4345dca49ba610ae8e3ce4955678c18c49c/google/cloud/firestore_v1/async_transaction.py#L324)
- [AsyncTransactional._pre_commit](https://github.com/googleapis/python-firestore/blob/9840d4345dca49ba610ae8e3ce4955678c18c49c/google/cloud/firestore_v1/async_transaction.py#L256)
- [Watch._on_snapshot_target_change_remove](https://github.com/googleapis/python-firestore/blob/9840d4345dca49ba610ae8e3ce4955678c18c49c/google/cloud/firestore_v1/watch.py#L404)

These should all use exception chaining to make sure the original error is preserved

daniel-sanche:
Previously, non-retryable exceptions raised during a transaction would be raised immediately, without causing a rollback. This PR adds an extra try...except block to make sure rollback is always called

It also uses exception chaining to resolve [actionable error issues](https://github.com/googleapis/python-firestore/issues/766) for the repo

Fixes: https://github.com/googleapis/python-firestore/issues/766, https://github.com/googleapis/python-firestore/issues/765


triplequark (approved):

google/cloud/firestore_v1/async_transaction.py
```diff
@@ -105,14 +105,18 @@ async def _begin(self, retry_id: bytes = None) -> None:
         )
         self._id = transaction_response.transaction
 
-    async def _rollback(self) -> None:
+    async def _rollback(self, source_exc=None) -> None:
```
danieljbruce, line 108: It seems unfortunate that we have to add an additional parameter to this function. I am guessing the underscore prefix means it is private? If so, I guess it's not that big of a deal, but if this argument is only being used for that one case in the __call__ function then could we catch errors there caused by `await transaction._rollback(source_exc=exc)` and do `raise exc from source_exc` there instead?

daniel-sanche: I'm not worried about the new parameter because 1) this is a private function, and 2) it's an optional paramater, so existing behaviour would remain consistent

But that's a good point that after this refactor, the rollback only happens in one place. In this case, we can rely on built-in exception chaining instead of `raise from`, since the exception occurs while handling a different exception. I simplified the code to remove the `source_exc`

PR #768, opened 2023-09-20
chore: remove owlbot from required checks

daniel-sanche:
removes owlbot from required, as it only needs to run when tracked files are changed


parthea (changes requested):
It's not recommended to remove `Owlbot Post Processor` as a required check. See comments in https://github.com/googleapis/python-datastore/pull/484#pullrequestreview-1662964724. Also see https://github.com/googleapis/repo-automation-bots/issues/1714

Issue #773, opened 2023-10-03
mypy issues


daniel-sanche:
Currently, mypy is just testing the `google.cloud.firestore` module, which is mostly empty. If changed to test `google.cloud.firestore_v1`, errors come up. We should resolve all typing issues with the library

One option is to completely wipe the type annotations and start fresh, slowly adding them back over time

Or, we can fix it [one file at a time](https://www.scivision.dev/mypy-type-check-one-file/), as we touch different areas of the code

daniel-sanche:
update: we are working on a couple new features, but this is a high priority item once those are in

daniel-sanche:
fixed in https://github.com/googleapis/python-firestore/pull/984

PR #782, opened 2023-10-16
fix: improve AsyncQuery typing

### LINKED ISSUE #781: Small typing improvement for AsyncCollectionReference and AsyncQuery
Typing isn't working well when using `AsyncCollectionReference` and `AsyncQuery`. As an example, `where` method which is inherited on both cases, will "return" a `BaseQuery`.

Having an AsyncQuery:
Captura de pantalla 2023-10-16 a las 16 56 45

After using `where` the editor thinks it is a `BaseQuery`:
Captura de pantalla 2023-10-16 a las 16 57 21

And because of this the editor says I can't do "async stuff":
Captura de pantalla 2023-10-16 a las 16 58 07

Correct typing for `AsyncCollectionReference` and `AsyncQuery` methods.

 **Additional context**
I'm also providing a small PR with required changes.

sergiterupri:

Fixes #781 🦕

### DIFF
--- a/google/cloud/firestore_v1/base_query.py
+++ b/google/cloud/firestore_v1/base_query.py
@@ -47,6 +47,7 @@ from typing import (
     Optional,
     Tuple,
     Type,
+    TypeVar,
     Union,
 )
 
@@ -101,6 +102,8 @@ _NO_ORDERS_FOR_CURSOR = (
 _MISMATCH_CURSOR_W_ORDER_BY = "The cursor {!r} does not match the order fields {!r}."
 
 _not_passed = object()
+
+QueryType = TypeVar("QueryType", bound="BaseQuery")
 
 
 class BaseFilter(abc.ABC):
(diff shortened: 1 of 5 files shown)


daniel-sanche:
Thanks for putting this together. I noticed recently that the [typechecking test is currently not working properly](https://github.com/googleapis/python-firestore/issues/773), so the types declared in this library will have flaws until that is resolved

google/cloud/firestore_v1/base_query.py
```diff
@@ -756,8 +764,9 @@ def end_before(
         )
 
     def end_at(
-        self, document_fields_or_snapshot: Union[DocumentSnapshot, dict, list, tuple]
-    ) -> "BaseQuery":
+        self: QueryType,
```
daniel-sanche, line 767: nit: it shouldn't be necessary to include type annotations for self (mypy doesn't require it, and we typically don't include them in other GCP libraries). 

Is there a reason it's needed here?

daniel-sanche (approved):

PR #788, opened 2023-11-02
fix: order normalization with descending query

daniel-sanche:
This PR fixes a bug in order normalization logic, to be more in line [with nodejs](https://github.com/googleapis/nodejs-firestore/blob/5da7138abc15e99edc51fdfd25a70bf5eb0c97b9/dev/src/reference.ts#L1972):

- all equality operators should be skipped when normalizing (previously, "in" and "array_contains_any" would still be added)
- always use the last direction when normalizing a query, instead of adding "ASCENDING" sections

### DIFF
--- a/google/cloud/firestore_v1/base_query.py
+++ b/google/cloud/firestore_v1/base_query.py
@@ -91,7 +91,7 @@ _EQUALITY_OPERATORS = (
     _operator_enum.EQUAL,
     _operator_enum.ARRAY_CONTAINS,
     _operator_enum.ARRAY_CONTAINS_ANY,
-    _operator_enum.IN
+    _operator_enum.IN,
 )
 _BAD_OP_STRING = "Operator string {!r} is invalid. Valid choices are: {}."
 _BAD_OP_NAN_NULL = 'Only an equality filter ("==") can be used with None or NaN values'
@@ -876,7 +876,10 @@ class BaseQuery(object):
                 if isinstance(filter_.op, StructuredQuery.FieldFilter.Operator):
                     field = filter_.field.field_path
                     # skip equality filters and filters on fields already ordered
-                    if filter_.op not in _EQUALITY_OPERATORS and field not in order_keys:
+                    if (
+                        filter_.op not in _EQUALITY_OPERATORS
+                        and field not in order_keys
+                    ):
                         orders.append(self._make_order(field, direction))
             # add __name__ if not already in orders
             if "__name__" not in [order.field.field_path for order in orders]:
--- a/google/cloud/firestore_v1/base_query.py
+++ b/google/cloud/firestore_v1/base_query.py
@@ -868,7 +868,7 @@ class BaseQuery(object):
                 _has_snapshot_cursor = True
         if _has_snapshot_cursor:
             # added orders should use direction of last order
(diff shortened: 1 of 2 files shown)


tests/unit/v1/test_base_query.py
```diff
@@ -1206,6 +1206,51 @@ def test_basequery__normalize_orders_w_name_orders_w_none_cursor():
...
+        .order_by("c", "DESCENDING")
+    query_w_snapshot = query.start_after(snapshot)
+
+    normalized = query._normalize_orders()
+    expected = [query._make_order("c", "DESCENDING")]
+    assert normalized == expected
+
+    normalized_w_snapshot = query_w_snapshot._normalize_orders()
+    expected_w_snapshot = expected + [
+        query._make_order("b", "DESCENDING"),
```
daniel-sanche, line 1250: let me know if this test case looks right. 

Previously we were always using ASCENDING for all added orders. My understanding is we should always use the previous direction?

google/cloud/firestore_v1/base_query.py
```diff
@@ -858,28 +866,24 @@ def _normalize_orders(self) -> list:
...
-            if not orders:
-                orders.append(self._make_order("__name__", "ASCENDING"))
-            else:
-                order_keys = [order.field.field_path for order in orders]
-                if "__name__" not in order_keys:
-                    direction = orders[-1].direction  # enum?
-                    orders.append(self._make_order("__name__", direction))
+                    # skip equality filters and filters on fields already ordered
+                    if (
+                        filter_.op not in _EQUALITY_OPERATORS
```
NickChittle, line 880: In my opinion, it would be better to create a list of INEQUALITY filters such that a new filter is default opted out rather than default opted in. This matches how other SDKs handle this more closely. I also think it's more likely someone will notice that an inequality is not having the order-by normalized rather than an equality filter is accidentally sneaking into the order-by normalization (Since it's an equality, the ordering is not easily noticed in most situations.)

daniel-sanche: makes sense, fixed

tests/unit/v1/test_base_query.py
NickChittle: Nit: IMO it would be more readable just to spell out the whole expected list here instead of appending the previous `expected`.

tests/unit/v1/test_base_query.py
```diff
@@ -1207,6 +1207,60 @@ def test_basequery__normalize_orders_w_name_orders_w_none_cursor():
...
+    """
+    from google.cloud.firestore_v1.base_query import FieldFilter
+
+    collection = _make_collection("here")
+    snapshot = _make_snapshot(_make_docref("here", "doc_id"), {"a": 1, "b": 2})
+    query = (
+        _make_base_query(collection)
+        .where(filter=FieldFilter("a", "==", 1))
+        .where(filter=FieldFilter("b", "in", [1, 2, 3]))
+        .where(filter=FieldFilter("b", "not-in", [4, 5, 6]))
```
NickChittle, line 1247: Nit: We have two 'b' values here, an equality and inequality on the same field, it would probably make more sense to have this be a separate property value.

NickChittle (approved):

danieljbruce (approved):

PR #843, opened 2024-02-16
feat: find emulator project id from environment variable

### LINKED ISSUE #174: Firestore Emulator Cloud Trigger Issue with Write Operations
Invoking an `update` or `set` operation to write to a Firestore document via the Firestore client does not work when called in a Firestore Emulator Cloud Trigger. Everything works fine when the function is deployed on the cloud.

In the following code example, whenever a document in `collection_A` gets updated, an event trigger does the ff:
- `field_1a` is set to `true` by using the event's document reference (WORKS)
- `field_2a` is set to `true` by using the Firestore client **(DOES NOT WORK)**
- Create a new document using the Firestore client **(DOES NOT WORK)**

```
from firebase_admin import firestore
from firebase_functions.firestore_fn import (Change, DocumentSnapshot, Event,
                                             on_document_updated)
from google.cloud.firestore import Client


@on_document_updated(document='collection_A/{document_id}')
def on_document_A_updated(event: Event[Change[DocumentSnapshot]]):
    new_doc = event.data.after
    data = new_doc.to_dict()

    if not data.get('field_1a'):
        # WORKS
        #
        # Using the document reference from the event works fine to update the
        # document.

        new_doc.reference.update({'field_1a': True})

    firestore_client: Client = firestore.Client()

    if not data.get('field_2a'):
        # DOES NOT WORK
        #
        # Using a document reference fetched from the client does not work
        # to update the document.

        firestore_client\
            .document(new_doc.reference.path)\
            .update({'field_2a': True})

    # DOES NOT WORK
    #
    # Creating a new document does not work.

    firestore_client\
        .collection('collection_B')\
        .document()\
        .set({'field_1b': 1})
```

daniel-sanche:
Previously, if `FIRESTORE_EMULATOR_HOST` was set and a `Client` instance was created without an explicit project, it would fall-back to `google-cloud-firestore-emulator`

This change makes it check the `GCLOUD_PROJECT` variable for a project id first, which is used by the firebase functions emulator. This means the client will auto-detect the project id if not given in the emulator, just like in a live GCP environment

Fixes https://github.com/firebase/firebase-functions-python/issues/174

BEGIN_COMMIT_OVERRIDE
fix: find emulator project id from environment variable
END_COMMIT_OVERRIDE

### DIFF
--- a/google/cloud/firestore_v1/base_client.py
+++ b/google/cloud/firestore_v1/base_client.py
@@ -132,7 +132,11 @@ class BaseClient(ClientWithProject):
                 credentials = AnonymousCredentials()
             if project is None:
                 # extract project from env var, or use system default
-                project = os.getenv("GCLOUD_PROJECT", _DEFAULT_EMULATOR_PROJECT)
+                project = (
+                    os.getenv("GOOGLE_CLOUD_PROJECT")
+                    or os.getenv("GCLOUD_PROJECT")
+                    or _DEFAULT_EMULATOR_PROJECT
+                )
 
         super(BaseClient, self).__init__(
             project=project,
--- a/google/cloud/firestore_v1/base_client.py
+++ b/google/cloud/firestore_v1/base_client.py
@@ -131,7 +131,8 @@ class BaseClient(ClientWithProject):
             if credentials is None:
                 credentials = AnonymousCredentials()
             if project is None:
-                project = _DEFAULT_EMULATOR_PROJECT
+                # extract project from env var, or use system default
+                project = os.getenv("GCLOUD_PROJECT", _DEFAULT_EMULATOR_PROJECT)
 
         super(BaseClient, self).__init__(
             project=project,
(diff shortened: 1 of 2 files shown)


google/cloud/firestore_v1/base_client.py
```diff
@@ -131,7 +131,8 @@ def __init__(
             if credentials is None:
                 credentials = AnonymousCredentials()
             if project is None:
-                project = _DEFAULT_EMULATOR_PROJECT
+                # extract project from env var, or use system default
+                project = os.getenv("GCLOUD_PROJECT", _DEFAULT_EMULATOR_PROJECT)
```
lahirumaramba, line 135: I think we should also add `GOOGLE_CLOUD_PROJECT` here

lahirumaramba (approved):
Thank you!! LGTM with adding support for `GOOGLE_CLOUD_PROJECT`!

gkevinzheng (approved):

PR #848, opened 2024-02-20
feat: Support `async` callbacks for `snapshot_callback()` function - `Watch` class

### LINKED ISSUE #836: feat: Support for `async` - `on_snapshot` callback
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

I wanted to run async callbacks when a new change in a given collection was occurring.

The solution I propose is to support async functions as callbacks as well, using asyncio, so developer using this won't notice the difference.

 **Describe alternatives you've considered**
In `google/cloud/firestore_v1/watch.py` - `push()` we could support async function as callback making use of `asyncio.run()`. I leave here the changes I made to make it work for myself. I added a single test to show the idea behind it.
[Potential Changes to propose in a PR](https://github.com/y3rbiadit0/python-firestore-async-on-snapshot-callback/tree/main)

y3rbiadit0:

Fixes  #836  🦕

# Context

I was working on a tiny project for myself and find out that I could need to run `async callbacks` when using `DocumentReference.on_snapshot()` function.

# Changes

1.  Added support for async callbacks in `push()` method (Line 540) in `Watch` class. 
2. Added `test_watch_async_callback` in `test_watch.py`
3. Added `custom_snapshot` arg to `_make_watch_no_mocks` - `test_watch.py` to allow mocking async callbacks.


daniel-sanche:
Hey, thanks for putting this together, but unfortunately it's not safe to use `asyncio.run` to run arbitrary coroutines as part of an otherwise synchronous function like this:

> [This function cannot be called when another asyncio event loop is running in the same thread.](https://docs.python.org/3/library/asyncio-runner.html#asyncio.run)

I saw you opened https://github.com/googleapis/python-firestore/issues/836, so I will use that thread to discuss alternatives. Thanks

PR #901, opened 2024-04-03
feat: support async Vector Search

pl04351820:
- [X] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/python-firestore/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [X] Ensure the tests and linter pass
- [X] Code coverage does not decrease (if any source code was changed)
- [X] Appropriate docs were updated (if necessary)

The PR supports async vector search for Firestore python.


tests/unit/v1/test_async_vector_query.py
```diff
@@ -0,0 +1,241 @@
...
+    # Verify the mock call.
+    parent_path, _ = parent._parent_info()
+    firestore_api.run_query.assert_called_once_with(
+        request={
+            "parent": parent_path,
+            "structured_query": async_vector_query._to_protobuf(),
+            "transaction": None,
+        },
+        metadata=client._rpc_metadata,
+    )
```
daniel-sanche, line 241: Could you add some system tests as well?

google/cloud/firestore_v1/async_vector_query.py
```diff
@@ -0,0 +1,127 @@
...
+        Returns:
+            list: The vector query results.
+        """
+        stream_result = self.stream(
+            transaction=transaction, retry=retry, timeout=timeout
+        )
+        result = [snapshot async for snapshot in stream_result]
+        return result  # type: ignore
+
+    async def stream(
```
daniel-sanche, line 76: I just realized `stream` isn't present in `BaseVectorQuery`, but it is implemented in both `VectorQuery` and `AsyncVectorQuery`. Is that intentional?

tests/unit/v1/test_async_vector_query.py
pl04351820: SG. Done.

google/cloud/firestore_v1/async_vector_query.py
pl04351820: Good point. Added one.

tests/system/test_system_async.py
```diff
@@ -339,6 +341,47 @@ async def test_document_update_w_int_field(client, cleanup, database):
     assert snapshot1.to_dict() == expected
 
 
+@pytest.mark.skipif(FIRESTORE_EMULATOR, reason="Require index and seed data")
```
daniel-sanche, line 344: Are you sure this is required? Are there no other system tests that would require an index?

daniel-sanche (approved):

tests/system/test_system_async.py
pl04351820: Yeah this is the only one I'm aware of. All others are depending on built-in index (no need to bootstrap).

PR #926, opened 2024-06-25
feat: use generator for stream results

Linchin:
- add an `StreamGenerator` class for all class methods `stream()` to return, instead of using the Python built-in generator.
- use `AsyncStreamGenerator` for the async clients


google/cloud/firestore_v1/query.py
```diff
@@ -506,3 +566,122 @@ def get_partitions(
             start_at = cursor
 
         yield QueryPartition(self, start_at, None)
+
+
+class Iterator(page_iterator.Iterator):
```
daniel-sanche, line 571: Why turn this into a page_iterator? Previously, it just returned a standard python generator

google/cloud/firestore_v1/query.py
```diff
@@ -386,6 +391,61 @@ def stream(
...
+        return Iterator(
+            self,
+            self._parent,
+            client,
+            transaction=transaction,
+            limit=limit,
+            retry=retry,
+            timeout=timeout,
+            all_descendants=self._all_descendants,
+        )
```
daniel-sanche, line 447: You could implement this as:

```
def stream(*args, **kwargs):
      inner_generator = self._old_stream(*args, **kwargs)
      return Iterator(inner_generator) 
```

And then the new custom Iterator would just wrap the old one:

```
class Iterator:

    def __init__(response_generator):
        self._generator = response_generator

    def __iter__():
        return self._response_generator
```

google/cloud/firestore_v1/query.py
daniel-sanche, line 571: also: I know this is early, but I'd probably prefer to give the iterator a more descriptive name

Linchin: I was using datastore as a reference, but apparently as you noted below, I could just use a class to wrap this method.

Linchin: I tentatively changed the name to `StreamIterator`. Maybe we can find a better name, such as `DocsIterator`?

google/cloud/firestore_v1/query.py
Linchin: This is a much simpler way to make an iterator! I updated the code to use this method.

google/cloud/firestore_bundle/bundle.py
```diff
@@ -222,7 +222,8 @@ async def _process_async_query(
     ) -> datetime.datetime:
         doc: DocumentSnapshot
         _read_time = datetime.datetime.min.replace(tzinfo=UTC)
-        async for doc in snapshot.stream():
+        docs = await snapshot.stream()
+        async for doc in docs:
```
daniel-sanche, line 226: we shouldn't have that extra await - the goal should be for this to be invisible to existing users

You should change `async def stream` to `def stream`, so it remains an AsyncGeneratorFunction, instead of becoming a Coroutine

google/cloud/firestore_v1/async_query.py
```diff
@@ -264,7 +269,7 @@ def avg(
         """
         return AsyncAggregationQuery(self).avg(field_ref, alias=alias)
 
-    async def stream(
+    async def _stream(
```
daniel-sanche, line 272: make rename this to _make_stream?

google/cloud/firestore_v1/async_query.py
```diff
@@ -412,3 +460,18 @@ async def get_partitions(
...
+
+class AsyncStreamIterator(abc.AsyncIterator):
+
+    def __init__(self, response_generator):
+        self._generator = response_generator
+
+    def __aiter__(self):
+        return self._generator
+    
+    def __anext__(self):
```
daniel-sanche: I think you shouldn't need to implement the next functions, only iter/aiter. But I could be wrong

If we do need __anext__ as well, you can probably just do
```
def __anext__(self):
  return self._generator.__anext__()
```

google/cloud/firestore_bundle/bundle.py
daniel-sanche: I don't think I explained this very well when we talked, so here's another try: 

- When `async` is used in the signature of a generator function, it tells Python to build an `AsyncGenerator` class instead of a standard `Generator`
- when `async` is used in standard function signatures, it says to wrap everything in a `Coroutine` that needs to be awaited
- We broke out the async generator into a helper `_stream` method. But we still want `_stream` to return an AsyncGenerator, not a coroutine. So we strip the `async` from the method signature, and turn it into a regular method. But it still behaves the same was as before

Note: this is also how the gapic generator generates [async streaming rpcs](https://github.com/googleapis/python-firestore/blob/9521deddc5a4b16956f37136f84928ac99688022/google/cloud/firestore_v1/services/firestore/async_client.py#L1182). They're standard functions that return async objects

google/cloud/firestore_v1/stream_iterator.py
```diff
@@ -0,0 +1,33 @@
...
+        self._generator = response_generator
+
+    def __iter__(self):
+        return self._generator
+    
+    def __next__(self):
+        try:
+            return next(self._generator)
+        except StopIteration:
+            return None
```
daniel-sanche, line 33: One thought: previously, we were returning generator objects. If you decide to implement __next__, maybe we should turn this into a full generator and implement __send__ and __throw__ too to be safe?

(I don't think these would be used by current users, but it would keep everything consistent)

google/cloud/firestore_bundle/bundle.py
Linchin: Thanks! Reverted these changes.

google/cloud/firestore_v1/async_query.py
Linchin: done

google/cloud/firestore_v1/async_query.py
Linchin: Thanks, this looks much more succinct! I think we will need to keep the next methods, because it's one of the methods that generator has.

google/cloud/firestore_v1/stream_iterator.py
Linchin: I feel like I had better implement these, because **send** and **throw** are quite useful in coroutines.

Linchin: If we were to go this route, should we rename the iterator to generator? Are we going too far on this 😂

google/cloud/firestore_v1/stream_iterator.py
```diff
@@ -0,0 +1,37 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Classes for iterating over stream results for the Google Cloud Firestore API.
+"""
+
+from collections import abc
+
+
+class StreamIterator(abc.Iterator):
```
daniel-sanche, line 21: as discussed in a different thread, let's just make this a generator to be safe

google/cloud/firestore_v1/stream_iterator.py
daniel-sanche: Yeah, I guess since the old type annotation was Generator, maybe we should make this into a full `StreamGenerator` class to be consistent (not that the type annotations can be relied on)

daniel-sanche (changes requested):
Mostly small suggestions. But lets make it into a generator to be safe

tests/unit/v1/test_async_stream_iterator.py
```diff
@@ -0,0 +1,82 @@
...
+    async def _inner_generator():
+        for i in iterable:
+            X = yield i
+            if X:
+                yield X
+
+    return AsyncStreamIterator(_inner_generator())
+
+
+@pytest.mark.skipif(sys.version_info < (3, 10), reason="requires python3.10 or higher")
```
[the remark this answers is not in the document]
Linchin: Because built-in `anext()` and `aiter()` were introduced in python 3.10 ([link](https://stackoverflow.com/a/73074973/8216546)). But on a second thought, I think I can call the dunder method directly, so we don't skip tests based on python version.

google/cloud/firestore_v1/collection.py
```diff
@@ -196,10 +195,10 @@ def get(
...
     def stream(
         self,
-        transaction: Union[Transaction, None] = None,
-        retry: retries.Retry = gapic_v1.method.DEFAULT,
-        timeout: Union[float, None] = None,
-    ) -> Generator[document.DocumentSnapshot, Any, None]:
+        transaction: Optional[transaction.Transaction] = None,
+        retry: Optional[retries.Retry] = gapic_v1.method.DEFAULT,
+        timeout: Optional[float] = None,
+    ) -> stream_iterator.StreamIterator:
```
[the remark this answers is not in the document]
Linchin: done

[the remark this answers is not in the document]
Linchin: indeed. Done as well.

google/cloud/firestore_v1/async_aggregation.py
```diff
@@ -150,13 +150,13 @@ def stream(
                 to a system-specified value.
 
         Returns:
-            async_stream_iterator.AsyncStreamIterator: A generator of the query
-            results.
+            `AsyncStreamGenerator[DocumentSnapshot]`:
```
daniel-sanche, line 153: Do you need the ` quotes here?

google/cloud/firestore_v1/async_aggregation.py
```diff
@@ -129,7 +129,7 @@ def stream(
         transaction: Optional[transaction.Transaction] = None,
         retry: Optional[retries.AsyncRetry] = gapic_v1.method.DEFAULT,
         timeout: Optional[float] = None,
-    ) -> async_stream_iterator.AsyncStreamIterator:
+    ) -> "AsyncStreamGenerator[DocumentSnapshot]":
```
daniel-sanche, line 132: Do you need quotes here? It looks like AsyncStreamGenerator is imported

google/cloud/firestore_v1/async_aggregation.py
```diff
@@ -25,14 +25,14 @@
...
-
+from google.cloud.firestore_v1.async_stream_generator import AsyncStreamGenerator
 from google.cloud.firestore_v1.base_aggregation import (
     AggregationResult,
     _query_response_to_result,
     BaseAggregationQuery,
 )
-
-from google.cloud.firestore_v1 import async_stream_iterator, transaction
+from google.cloud.firestore_v1.base_document import DocumentSnapshot
```
daniel-sanche, line 34: Is this just used for type annotations? If so, you can put it in an [`if TYPE_CHECKING` block](https://github.com/googleapis/python-bigtable/blob/c573e9b695658f3e0c78eb3443d0cf26fea57da9/google/cloud/bigtable/data/_async/_read_rows.py#L42) to avoid importing it in production

daniel-sanche (approved):
A couple small comments about imports and docstrings, but LGTM!

google/cloud/firestore_v1/async_aggregation.py
Linchin: I added the quotes because `[DocumentSnapshot]` part would cause an error, because `AsyncStreamGenerator` isn't recognized as a generator in type hints. I wasn't able to find how to do it, so I resorted to using a string. I don't have a strong opinion on this. What do you think?

google/cloud/firestore_v1/async_aggregation.py
Linchin: Honestly I'm not super sure about this. I added the backticks because they are used in many other type annotations (for [example](https://github.com/googleapis/python-firestore/blob/104293b88cfc0af411136d87bdac67810cb938fb/google/cloud/firestore_v1/client.py#L170)). (A quick search shows the backtick makes sphinx [create a link in the docs](https://developer.lsst.io/v/u-ktl-third-party-team/python/numpydoc.html#marking-up-parameter-names), but I'm not very clear about which documentation standard we are following.) Maybe we can clean this up along with other type annotations in a future PR. WDYT?

PR #938, opened 2024-07-17
feat: query profiling part 1: synchronous

### LINKED ISSUE #947: type issue: "StreamGenerator" is not iterable
This issue was introduced with the changes in 2.17.0. It exists for each of the pyright version I tested at random.

It is only a type issue -- the code behaves as expected.

#### Environment details

  - OS type and version: MacOS
  - Python version: `python --version` 3.11.7
  - pip version: `pip --version` poetry
  - `google-cloud-firestore` version: `pip show google-cloud-firestore` 2.17.0
  - `pyright` version: 1.1.375

#### Steps to reproduce

  1. Setup environment as above
  2. Run pyright

#### Code example

```python
client = firestore_v1.Client()


def example():
    collection = client.collection("foo")
    docs = collection.stream()
    for doc in docs:
        print(doc)
```

image

#### Error message
```
eample.py
  example.py:7:20 - error: "StreamGenerator" is not iterable
    "__next__" method not defined on type "DocumentSnapshot" (reportGeneralTypeIssues)
```

Linchin:
The Synchronous section for query profiling support.

Checklist:
- [x] add the class definition query_profile.py
- [x] add class imports in init.py
- [x] methods:
  - [x] query.stream, .get
  - [x] collection.stream, .get  -- calls query
  - [ ] ~~CollectionGroup.stream, .get, -- inherits query~~
  - [x] AggregationQuery.stream, .get
  - [x] vector.stream, .get 
  - [x] transaction.get
  - [ ] ~~client.get_all~~
- [x] add system test for transaction
- [x] stream generator unit tests
- [x] DocumentSnapshotList class
- [x] DocumentSnapshotList unit tests
- [x] system test for each class
  - [x] aggregation
  - [x] collection
  - [ ] ~~document~~ only get, and does not use the stream api
  - [x] vector_query
  - [x] query
  - [x] transaction

Also closes #947


daniel-sanche (commented):
Leaving a couple initial thoughts, but still looking through things

daniel-sanche:
We would benefit from some time verification here, but the mypy check is [currently broken](https://github.com/googleapis/python-firestore/issues/773). Maybe you should try to enable it just for the new classes/code you touch? I don't know how practical that would be though

google/cloud/firestore_v1/aggregation.py
```diff
@@ -54,6 +58,7 @@ def get(
             retries.Retry, None, gapic_v1.method._MethodDefault
         ] = gapic_v1.method.DEFAULT,
         timeout: float | None = None,
+        explain_options: Optional[ExplainOptions] = None,
```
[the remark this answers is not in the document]
daniel-sanche: Also, probably better to follow existing style (from the timeout argument) and use `ExplainOptions | None` instead of `Optional`

google/cloud/firestore_v1/base_document.py
```diff
@@ -503,6 +504,48 @@ def _to_protobuf(self) -> Optional[Document]:
...
+        explain_options: Optional[query_profile.ExplainOptions] = None,
+        explain_metrics: Optional[query_profile.ExplainMetrics] = None,
+    ):
+        super().__init__(docs)
+        self._explain_options = explain_options
+        self._explain_metrics = explain_metrics
+
+    @property
+    def explain_options(self):
+        return self._explain_options
```
daniel-sanche, line 539: Does StreamGenerator also expose `explain_options`?

google/cloud/firestore_v1/base_transaction.py
```diff
@@ -20,6 +20,9 @@
 
 from google.cloud.firestore_v1 import types
 
+# Types needed only for Type Hints
+from google.cloud.firestore_v1.query_profile import ExplainOptions
```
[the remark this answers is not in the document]
Linchin: Indeed, I have added the type checking flag.

google/cloud/firestore_v1/aggregation.py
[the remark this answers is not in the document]
Linchin: > Can you add a * above `explain_options

Thanks for pointing this out, it's great to know! Added.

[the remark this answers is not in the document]
Linchin: > Also, probably better to follow existing style (from the timeout argument) and use ExplainOptions | None instead of Optional

I'm not quite sure about this, because the type hint style is a mix right now. Maybe we can save it for a docstring overhaul later, when we are certain about which style we will go for?

google/cloud/firestore_v1/base_document.py
```diff
@@ -503,6 +504,48 @@ def _to_protobuf(self) -> Optional[Document]:
...
+class DocumentSnapshotList(list):
+    """A list of DocumentSnapshot.
+
+    This is a subclass of the built-in list. A new property `explain_metrics`
+    is added to return the query profile results.
+
+    Args:
+        reference (:class:`~google.cloud.firestore_v1.document.DocumentReference`):
+            A document reference corresponding to the document that contains
+            the data in this snapshot.
```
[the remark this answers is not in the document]
Linchin: yes, removed

google/cloud/firestore_v1/aggregation.py
```diff
@@ -70,20 +75,37 @@ def get(
...
+            explain_options=explain_options,
+        )
+        result_list = list(result)
 
-    def _get_stream_iterator(self, transaction, retry, timeout):
+        if explain_options is None:
+            explain_metrics = None
+        else:
+            explain_metrics = result.explain_metrics
+
```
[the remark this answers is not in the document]
Linchin: Indeed this same if/else code appears repeatedly in all get() methods. But considering we want to expand DocumentSnapshotList into a more general list class that may encompass types other than DocumentSnapshot, I wonder if the same logic checking explain_options will apply to all future classes that form the list? Also if we put this logic inside DocumentSnapshotList, we would need to pass the original `result` (StreamGenerator) into the initialization. I personally feel like it's cleaner to initialize with the explain_metrics, but I also want to hear your opinion on this.

google/cloud/firestore_v1/base_document.py
```diff
@@ -503,6 +504,48 @@ def _to_protobuf(self) -> Optional[Document]:
...
+    @property
+    def explain_options(self):
+        return self._explain_options
+
+    @property
+    def explain_metrics(self):
+        if self._explain_options is None:
+            raise query_profile.QueryExplainError("explain_options not set on query.")
+        else:
+            return self._explain_metrics
```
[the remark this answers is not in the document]
Linchin: I agree this is repeating code, but also I feel like using mixin for one property might be a bit of overkill? I think you have a good point that we are raising a lot of exceptions with exactly the same error message. Would it make sense to create a few subclasses of QueryExplainError for each of the cases, so at least we have a central place to update the error messages?

google/cloud/firestore_v1/base_document.py
Linchin: Good catch! Added the property.

google/cloud/firestore_v1/aggregation.py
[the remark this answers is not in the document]
daniel-sanche: Yeah that works for me!

google/cloud/firestore_v1/stream_generator.py
```diff
@@ -38,3 +65,35 @@ def throw(self, exp=None):
...
+            return self._explain_metrics
+        elif self._explain_options is None:
+            raise QueryExplainError("explain_options not set on query.")
+        elif self._explain_options.analyze is False:
+            # we need to run the query to get the explain_metrics
+            try:
+                next(self)
+            except StopIteration:
+                pass
+            return self._explain_metrics
```
daniel-sanche, line 96: there is a problem with the typing here: `self._explain_metrics` is optional, but this method says it always returns an `ExplainMetrics`. Maybe raise an exception if it's still None after calling next? Or is there a cleaner way?

google/cloud/firestore_v1/stream_generator.py
```diff
@@ -38,3 +65,35 @@ def throw(self, exp=None):
...
+            if explain_metrics is not available on the query.
+        """
+        if self._explain_metrics is not None:
+            return self._explain_metrics
+        elif self._explain_options is None:
+            raise QueryExplainError("explain_options not set on query.")
+        elif self._explain_options.analyze is False:
+            # we need to run the query to get the explain_metrics
+            try:
+                next(self)
```
daniel-sanche, line 93: This structure is going to cause issues when you make the async version. `@property` methods can't call coroutines.

Maybe we should just raise an exception if this is accessed before it's iterated? Or do you have another idea?

google/cloud/firestore_v1/aggregation.py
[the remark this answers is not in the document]
daniel-sanche: On second thought, I think you're right, it's probably better to leave most of this stuff outside the __init__

I was thinking it could be cleaner to just pass in a `StreamGenerator` on __init__, since it should contain all the information needed by `DocumentSnapshotList`. But that gets complicated on the async side, because __init__ isn't an async function. So we'd have trouble building the object there

Still though, one thing to consider: if we pass in the `StreamGenerator` and hold a reference to it in `DocumentSnapshotList`, you can delegate to that instance when calling `explain_metrics()`, which could help reduce some duplication

google/cloud/firestore_v1/stream_generator.py
```diff
@@ -14,21 +14,48 @@
...
+from google.cloud.firestore_v1.query_profile import ExplainMetrics, QueryExplainError
+import google.cloud.firestore_v1.types.query_profile as query_profile_proto
+
+if TYPE_CHECKING:  # pragma: NO COVER
+    from google.cloud.firestore_v1.query_profile import ExplainOptions
 
 
 class StreamGenerator(abc.Generator):
-    """Generator for the streamed results."""
+    """Generator for the streamed results.
```
daniel-sanche, line 30: Please use generic for the typing here

google/cloud/firestore_v1/base_document.py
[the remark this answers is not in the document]
daniel-sanche: Yeah, I think you're right, Mixins probably only make sense if the shared code starts to expand, so maybe just keep it in mind if it seems it may be useful at some point. Probably overkill for now

One other thought I mentioned in a different thread though: DocumentSnapshotList could hold an instance of StreamGenerator, and delegate this call to the inner stream instance, which would be another way to reduce duplication

google/cloud/firestore_v1/stream_generator.py
daniel-sanche, line 93: Also, wouldn't the value from `next()` be lost here?

Linchin: I think when `analyze == True`,  the query isn't run so there's no query results returned anyway. `explain_metrics` is parsed when calling `next()`. Are there any other returned value that I'm not aware of?

google/cloud/firestore_v1/aggregation.py
```diff
@@ -154,6 +182,10 @@ def _make_stream(
 
             if response is None:  # EOI
                 break
+
+            if response.explain_metrics:
+                yield response.explain_metrics
```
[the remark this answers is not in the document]
Linchin: Good point! Done

google/cloud/firestore_v1/base_document.py
```diff
@@ -503,6 +504,48 @@ def _to_protobuf(self) -> Optional[Document]:
...
+            (Optional[:class:`~google.cloud.firestore_v1.query_profile.ExplainOptions`]):
+            Options to enable query profiling for this query. When set,
+            explain_metrics will be available on the returned generator.
+        explain_metrics (Optional[query_profile.ExplainMetrics]):
+            Query profile results.
+    """
+
+    def __init__(
+        self,
+        docs: list,
```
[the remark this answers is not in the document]
Linchin: I feel like `list[DocumentSnapshot]` doesn't indicate that the result has more methods than `list`, so I'm more prone to `QueryResults` or `QueryResultsList`.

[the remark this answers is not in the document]
daniel-sanche: Sorry, I think I was talking about two different things in these comments:
- the class itself should be called either `QueryResults[T]` or `QueryResultsList[T]` (I kind of lean towards QueryResultsList, but I'm fine with either)
- the `docs: list` init argument should have a generic component to it: `docs: list[T]`

google/cloud/firestore_v1/base_document.py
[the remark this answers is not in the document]
Linchin: Great idea! I will add it along with the other list related changes

google/cloud/firestore_v1/stream_generator.py
Linchin: Indeed, I think it's better to add an exception if `explain_metrics` is empty. I'm a bit curious as to when this can happen, maybe the backend didn't send the proper data?

google/cloud/firestore_v1/stream_generator.py
Linchin: > Maybe we should just raise an exception if this is accessed before it's iterated? Or do you have another idea?

With `explain_options.analyze == False`, the query isn't really run, and there's no query result returned. Thus users won't organically iterate the results themselves. I'm not the most knowledgeable person about async python, but it seems possible to make an async property, but it's not recommended isn't very convenient to use. (see [stackoverflow thread](https://stackoverflow.com/questions/54984337/how-should-you-create-properties-when-using-asyncio).) 

Apart from forcing an async property, other options I can think of:
1. Move the logic to inside `Query._make_stream()` ([here](https://github.com/googleapis/python-firestore/blob/a0d2d7e26bd2008cc1f19727002a8c663bce78ec/google/cloud/firestore_v1/query.py#L362)): if `analyze == False`, get `explain_metrics` right after iterator is created, so StreamGenerator doesn't need to call the API anymore. This may make creating the generator a bit slower, though.
2. make explain_metrics not a property but a function. At least formally it's not a property anymore - but this will create a difference between datastore and firestore SDKs

I feel like (1) is a better idea. WDYT?

google/cloud/firestore_v1/stream_generator.py
Linchin: done

google/cloud/firestore_v1/base_document.py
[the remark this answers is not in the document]
Linchin: I like QueryResultsList more too! done

google/cloud/firestore_v1/aggregation.py
```diff
@@ -54,6 +58,7 @@ def get(
             retries.Retry, None, gapic_v1.method._MethodDefault
         ] = gapic_v1.method.DEFAULT,
         timeout: float | None = None,
+        explain_options: Optional[ExplainOptions] = None,
     ) -> List[AggregationResult]:
```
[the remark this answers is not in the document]
Linchin: done

google/cloud/firestore_v1/stream_generator.py
Linchin: I actually copied from your PR. This is from your branch: https://github.com/googleapis/python-datastore/blob/e67ca728a982b15b6d2bc0f18fe9c27f07014471/google/cloud/datastore/query.py#L871-L872

Also updated the comments.

google/cloud/firestore_v1/aggregation.py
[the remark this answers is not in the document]
Linchin: When I'm trying to implement this, I still feel a bit unsure. Passing `StreamGenerator` feels like we are coupling it with `QueryResultsList` really deeply and may make these two classes less versatile in the future.

google/cloud/firestore_v1/stream_generator.py
Linchin: done

google/cloud/firestore_v1/collection.py
```diff
@@ -169,7 +171,8 @@ def get(
         transaction: Union[transaction.Transaction, None] = None,
         retry: retries.Retry = gapic_v1.method.DEFAULT,
         timeout: Union[float, None] = None,
-    ) -> list:
+        explain_options: Optional[ExplainOptions] = None,
+    ) -> DocumentSnapshotList:
```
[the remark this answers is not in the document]
Linchin: indeed, done

google/cloud/firestore_v1/stream_generator.py
daniel-sanche: I don't think 1 solves the problem. In that case, _make_stream will be an AsyncGenerator, and you'll need to await generator.__anext__ to get the explain_metrics value from it. (Or maybe I'm misunderstanding the details, but fundamentally, you'll have to await something somewhere to get that value, so we can't use a sync property)

I think 2 is cleaner. Turn it into `get_explain_metrics()`. I think it's ok that it's different from datastore here

Linchin: For 1, I meant we get `explain_metrics` when  `Query._make_stream()` is called, and pass the value to the init function of StreamGenerator, so `explain_metrics` will be strictly a property without any operation to get data when called

google/cloud/firestore_v1/aggregation.py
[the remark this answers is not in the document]
daniel-sanche: If it seems messy, don't worry about it, I was just throwing out ideas. We can just re-visit this if the duplicated code expands in the future

IMO I think it makes sense for `StreamGenerator` and `QueryResultsList` to be coupled in some way though, either through inheritance or composition. They have essentially the same use-case, QueryResultsList just has the stream fully loaded into memory

google/cloud/firestore_v1/base_document.py
```diff
@@ -503,6 +522,45 @@ def _to_protobuf(self) -> Optional[Document]:
         return _helpers.document_snapshot_to_protobuf(self)
 
 
+class QueryResultsList(list):
```
daniel-sanche, line 525: maybe we should move this in with stream_generator.py? Or it's own file?

It made sense to leave this here when it only held DocumentReferences, but not so much now that it's generic

google/cloud/firestore_v1/aggregation.py
```diff
@@ -127,16 +152,21 @@ def _make_stream(
...
             timeout (Optional[float]): The timeout for this request.  Defaults
                 to a system-specified value.
+            explain_options
+                (Optional[:class:`~google.cloud.firestore_v1.query_profile.ExplainOptions`]):
+                Options to enable query profiling for this query. When set,
+                explain_metrics will be available on the returned generator.
 
         Yields:
-            :class:`~google.cloud.firestore_v1.base_aggregation.AggregationResult`:
+            Tuple[Optional[DocumentSnapshot], Optional[ExplainMetrics]]:
```
daniel-sanche, line 161: shouldn't `Optional[DocumentSnapshot]` be `List[AggregationResult]`?

google/cloud/firestore_v1/aggregation.py
```diff
@@ -154,14 +184,20 @@ def _make_stream(
 
             if response is None:  # EOI
                 break
+
+            if response.explain_metrics:
+                yield None, response.explain_metrics
```
daniel-sanche, line 189: no need to yield here: you can yield them both together a couple lines later. That way, only the second value needs to be optional

```
maybe_metrics = response.explain_metrics or None
result = _query_response_to_result(response)
yield result, maybe_metrics
```

google/cloud/firestore_v1/base_aggregation.py
```diff
@@ -230,6 +234,8 @@ def get(
             retries.Retry, None, gapic_v1.method._MethodDefault
         ] = gapic_v1.method.DEFAULT,
         timeout: float | None = None,
+        *,
+        explain_options: Optional[ExplainOptions] = None,
     ) -> List[AggregationResult] | Coroutine[Any, Any, List[AggregationResult]]:
```
daniel-sanche, line 239: update the return type? (looks like there are a lot that can be changed across the files)

google/cloud/firestore_v1/transaction.py
```diff
@@ -171,6 +174,8 @@ def get(
         ref_or_query,
         retry: retries.Retry = gapic_v1.method.DEFAULT,
         timeout: float = None,
+        *,
+        explain_options: Optional[ExplainOptions] = None,
     ) -> Generator[DocumentSnapshot, Any, None]:
```
daniel-sanche, line 179: change type?

google/cloud/firestore_v1/stream_generator.py
daniel-sanche: I don't think we can, because `self._make_stream` is just making an AsyncGenerator when it is [called for the first time](https://github.com/googleapis/python-firestore/blob/cb693f8b02434de803ce41e72bb5aa18c5de3764/google/cloud/firestore_v1/async_query.py#L401C32-L401C44). Basically like an __init__ function. It's code isn't executed until __anext__ is called on it.

If AsyncQuery.stream were an async function (aka coroutine), then we could do something there. But as-is, we're not in an async context at any point there

google/cloud/firestore_v1/stream_generator.py
```diff
@@ -50,10 +49,14 @@ def __iter__(self):
...
-            self._explain_metrics = ExplainMetrics._from_pb(next_value)
-            return self._generator.__next__()
+        next_value, explain_metrics = self._generator.__next__()
+
+        if explain_metrics is not None:
+            self._explain_metrics = ExplainMetrics._from_pb(explain_metrics)
+
+            # Need to run the following iteration too, to ensure the length of
+            # the iteration isn't increased due to yielding explain_metrics.
+            return next(self)
```
daniel-sanche, line 62: We can simplify this if we change the stream to yield both values together, right?

google/cloud/firestore_v1/stream_generator.py
Linchin: Wow I see, didn't know that `AsyncQuery.stream()` isn't async. Then I will go with option 2.

google/cloud/firestore_v1/base_document.py
Linchin: True! moved

google/cloud/firestore_v1/aggregation.py
Linchin: indeed, updated

google/cloud/firestore_v1/base_aggregation.py
Linchin: done, also checked the whole file

google/cloud/firestore_v1/aggregation.py
Linchin: I like the idea of returning the value instead! This way we can avoid calling next() internally too.

google/cloud/firestore_v1/transaction.py
Linchin: done

google/cloud/firestore_v1/base_document.py
Linchin: I moved to its own file instead.

google/cloud/firestore_v1/stream_generator.py
```diff
@@ -14,21 +14,54 @@
...
+
+
+T = TypeVar("T")
 
 
 class StreamGenerator(abc.Generator):
-    """Generator for the streamed results."""
+    """Generator for the streamed results.
 
-    def __init__(self, response_generator):
```
daniel-sanche, line 24: FYI: https://github.com/googleapis/python-firestore/issues/947

Maybe we should look at this as part of the changes?

Linchin: Thanks, I think with the mypy change this should be resolved, although there are still many type annotation errors remaining.

google/cloud/firestore_v1/stream_generator.py
Linchin: done

google/cloud/firestore_v1/base_aggregation.py
```diff
@@ -62,7 +66,7 @@ class AggregationResult(object):
     :param value: The resulting read_time
     """
 
-    def __init__(self, alias: str, value: int, read_time=None):
+    def __init__(self, alias: str, value: int | float, read_time=None):
```
daniel-sanche, line 69: You can just use `float` instead of `int | float` (float implies int)

google/cloud/firestore_v1/base_aggregation.py
```diff
@@ -230,7 +237,12 @@ def get(
             retries.Retry, None, gapic_v1.method._MethodDefault
         ] = gapic_v1.method.DEFAULT,
         timeout: float | None = None,
-    ) -> List[AggregationResult] | Coroutine[Any, Any, List[AggregationResult]]:
+        *,
+        explain_options: Optional[ExplainOptions] = None,
+    ) -> (
+        QueryResultsList[AggregationResult]
+        | Coroutine[Any, Any, List[AggregationResult]]
```
daniel-sanche, line 244: Shouldn't the Coroutine also return QueryResultsList? (Also, `Awaitable[` seems cleaner than `Coroutine[Any, Any`

google/cloud/firestore_v1/base_aggregation.py
```diff
@@ -246,21 +258,30 @@ def get(
...
+            retries.Retry, None, gapic_v1.method._MethodDefault
+        ] = gapic_v1.method.DEFAULT,
         timeout: Optional[float] = None,
+        *,
+        explain_options: Optional[ExplainOptions] = None,
     ) -> (
-        Generator[List[AggregationResult], Any, None]
-        | AsyncGenerator[List[AggregationResult], None]
+        StreamGenerator[Optional[List[AggregationResult]]]
+        | AsyncStreamGenerator[List[AggregationResult], Any, None]
```
daniel-sanche, line 284: does AsyncStreamGenerator exist in this PR?

google/cloud/firestore_v1/base_document.py
```diff
@@ -13,17 +13,33 @@
...
 from google.cloud.firestore_v1 import field_path as field_path_module
+from google.cloud.firestore_v1.types import common
 
 # Types needed only for Type Hints
-from google.cloud.firestore_v1.types import Document, common, firestore, write
+if TYPE_CHECKING:  # pragma: NO COVER
+    from google.cloud.firestore_v1.types import Document, firestore, write
+
+
+T = TypeVar("T")
```
daniel-sanche, line 42: is this needed?

google/cloud/firestore_v1/base_vector_query.py
```diff
@@ -124,5 +134,7 @@ def stream(
         transaction=None,
         retry: retries.Retry = gapic_v1.method.DEFAULT,
         timeout: float = None,
-    ) -> Iterable[document.DocumentSnapshot]:
+        *,
+        explain_options: Optional[ExplainOptions] = None,
+    ) -> Generator[Optional[DocumentSnapshot], Any, Optional[ExplainMetrics]]:
```
daniel-sanche, line 139: why Optional[DocumentSnapshot]?

google/cloud/firestore_v1/query.py
```diff
@@ -203,7 +223,7 @@ def _chunkify(
             snapshots = _q.get()
 
             if snapshots:
-                last_document = snapshots[-1]
+                last_document = snapshots[-1]  # type: ignore
```
daniel-sanche, line 226: why type ignore?

google/cloud/firestore_v1/query.py
```diff
@@ -349,15 +371,26 @@ def _make_stream(
...
             timeout (Optional[float]): The timeout for this request. Defaults
                 to a system-specified value.
+            explain_options
+                (Optional[:class:`~google.cloud.firestore_v1.query_profile.ExplainOptions`]):
+                Options to enable query profiling for this query. When set,
+                explain_metrics will be available on the returned generator.
 
         Yields:
-            :class:`~google.cloud.firestore_v1.document.DocumentSnapshot`:
+            Optional[DocumentSnapshot]:
```
daniel-sanche, line 380: I don't think this needs to be optional

google/cloud/firestore_v1/stream_generator.py
```diff
@@ -14,21 +14,50 @@
...
+)
 
-class StreamGenerator(abc.Generator):
-    """Generator for the streamed results."""
+if TYPE_CHECKING:  # pragma: NO COVER
+    from google.cloud.firestore_v1.query_profile import ExplainOptions
 
-    def __init__(self, response_generator):
+
+class StreamGenerator(Generator[Any, Any, Optional[ExplainMetrics]]):
```
daniel-sanche, line 30: 1. We should use a generic for the output type, instead of Any
2. We probably don't need to return Optional[ExplainMetrics] to the end user in the API contract, we can treat that as an implementation detail. Users should access explain_metrics by calling the method

google/cloud/firestore_v1/aggregation.py
```diff
@@ -154,15 +193,29 @@ def _make_stream(
...
+
             result = _query_response_to_result(response)
-            yield result
+            if len(result) > 0 or metrics is None:
+                # When explain_options.analyze == False, the query isn't run
+                # but query profiling results are returned. In this case we
+                # do not yield an empty result.
+                yield result
+
+        return metrics
```
daniel-sanche, line 207: I'm realizing AsyncGenerator doesn't support returning values, so we'll have to find a different method for the async side... I don't know if that changes anything here

daniel-sanche (changes requested):

google/cloud/firestore_v1/base_vector_query.py
Linchin: I added the `Optional` to cover the case when the generator generates 0 items. Maybe it's not necessary?

google/cloud/firestore_v1/query_results.py
```diff
@@ -0,0 +1,74 @@
...
+    ExplainMetrics,
+    ExplainOptions,
+    QueryExplainError,
+)
+
+
+from typing import List, Optional, TypeVar
+
+
+T = TypeVar("T")
```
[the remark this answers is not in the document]
Linchin: Good catch! removed.

google/cloud/firestore_v1/aggregation.py
```diff
@@ -127,16 +155,27 @@ def _make_stream(
...
             timeout (Optional[float]): The timeout for this request.  Defaults
                 to a system-specified value.
+            explain_options
+                (Optional[:class:`~google.cloud.firestore_v1.query_profile.ExplainOptions`]):
+                Options to enable query profiling for this query. When set,
+                explain_metrics will be available on the returned generator.
 
         Yields:
-            :class:`~google.cloud.firestore_v1.base_aggregation.AggregationResult`:
+            Optional[List[AggregationResult]]:
```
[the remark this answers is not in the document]
Linchin: I think you are right, even without the Optional the generator should be allowed to be empty.

google/cloud/firestore_v1/aggregation.py
```diff
@@ -154,15 +193,29 @@ def _make_stream(
...
+            if metrics is None and response.explain_metrics:
+                metrics = response.explain_metrics
+
             result = _query_response_to_result(response)
-            yield result
+            if len(result) > 0 or metrics is None:
+                # When explain_options.analyze == False, the query isn't run
+                # but query profiling results are returned. In this case we
+                # do not yield an empty result.
+                yield result
```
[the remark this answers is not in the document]
Linchin: I think this logic was intended to cover the case when `analyze==True` but aggregation result is empty. In this case we still yield an empty list. I added more comments to explain the reasoning more clearly. I also changed the condition to be more direct, as following:
```
if len(result) > 0 or explain_options.analyze is True:
```

google/cloud/firestore_v1/aggregation.py
```diff
@@ -154,15 +193,29 @@ def _make_stream(
...
         transaction: Optional["transaction.Transaction"] = None,
-        retry: Optional[retries.Retry] = gapic_v1.method.DEFAULT,
+        retry: Union[
+            retries.Retry, None, gapic_v1.method._MethodDefault
+        ] = gapic_v1.method.DEFAULT,
         timeout: Optional[float] = None,
-    ) -> "StreamGenerator[DocumentSnapshot]":
+        *,
+        explain_options: Optional[ExplainOptions] = None,
+    ) -> StreamGenerator[Optional[List[AggregationResult]]]:
```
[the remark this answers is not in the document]
Linchin: Indeed, as you can see [here](https://github.com/googleapis/python-firestore/blob/ea64f23a68d7610ce7ed3369c723e2cbffd2c56d/google/cloud/firestore_v1/aggregation.py#L111) the internal `_make_stream()` has always yielded AggregationResult before.

google/cloud/firestore_v1/base_aggregation.py
Linchin: That's good to know! Replaced all occurrences.

google/cloud/firestore_v1/base_aggregation.py
Linchin: I didn't change this part because async isn't implemented in this PR. I will change it to use `Awaitable[` instead.

google/cloud/firestore_v1/base_aggregation.py
Linchin: I think `stream()` does return AsyncStreamGenerator, [see here](https://github.com/googleapis/python-firestore/blob/ea64f23a68d7610ce7ed3369c723e2cbffd2c56d/google/cloud/firestore_v1/async_aggregation.py#L164).

Linchin: But I removed the async return value :)

google/cloud/firestore_v1/base_document.py
Linchin: Thanks for catching this. I think I added generic type here at first, but removed because sometimes it's not accepted by python 3.7.

google/cloud/firestore_v1/query.py
Linchin: Some unit test wasn't passing but somehow it works now. I'll undo the change.

google/cloud/firestore_v1/query.py
Linchin: Indeed, done

google/cloud/firestore_v1/stream_generator.py
Linchin: I think you are right about both. I will update accordingly.

google/cloud/firestore_v1/aggregation.py
Linchin: I encountered this issue when implementing the async part. The temp hacky solution is to yield the metrics and catch it by checking type 😂 I think I really like the return here so I'd rather the sync and async aren't symmetric here, but let's discuss in more details in our meeting.

google/cloud/firestore_v1/base_vector_query.py
Linchin: I think it's not necessary. I removed all similar occurrences.

google/cloud/firestore_v1/base_document.py
Linchin: removed

google/cloud/firestore_v1/base_vector_query.py
daniel-sanche: Yeah, that's not necessary. Optional implies it would sometimes `yield None`, which would be different from just not yielding anything

google/cloud/firestore_v1/aggregation.py
daniel-sanche: Yeah, I guess we're back to our old worse options for the Async side! I agree it's probably fine to keep them separate if it saves us headaches on the sync side at least

google/cloud/firestore_v1/aggregation.py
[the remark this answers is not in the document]
daniel-sanche: That makes sense. What happens if we just don't yield anything though? Does it break when analyze is True?

google/cloud/firestore_v1/base_aggregation.py
```diff
@@ -230,7 +237,12 @@ def get(
             retries.Retry, None, gapic_v1.method._MethodDefault
         ] = gapic_v1.method.DEFAULT,
         timeout: float | None = None,
-    ) -> List[AggregationResult] | Coroutine[Any, Any, List[AggregationResult]]:
+        *,
+        explain_options: Optional[ExplainOptions] = None,
+    ) -> (
+        QueryResultsList[AggregationResult]
+        | Awaitable[Any, Any, List[AggregationResult]]
```
daniel-sanche, line 244: I think Awaitable only takes one argument. And I think it should be QueryResultsList, not just List

google/cloud/firestore_v1/stream_generator.py
```diff
@@ -14,21 +14,53 @@
...
+    ):
         self._generator = response_generator
+        self._explain_options = explain_options
+        self._explain_metrics = None
 
     def __iter__(self):
-        return self._generator
+        return self
 
     def __next__(self):
```
daniel-sanche, line 56: nit: you can probably add `-> T` return annotation here, which might help with type checking

google/cloud/firestore_v1/aggregation.py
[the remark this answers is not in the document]
daniel-sanche: It seems like we have to decide what we want to happen when analyze is True:
- yield a single, empty list
- just raise a StopIteration, as an empty iterator

If we yield an emtpy list, that means `list(genrator) == [[]]`. I think just having no yields feels cleaner to me. WDYT?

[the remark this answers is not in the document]
Linchin: Indeed yielding nothing didn't break any test, I'll just do this!

google/cloud/firestore_v1/stream_generator.py
Linchin: done

daniel-sanche:
One last thing: I ran mypy locally, and found a couple new type issues that seem relevant to these changes

```
tmp/python-firestore [query-profiling-3●] » mypy -p google.cloud.firestore_v1.base_aggregation                                                                                                                                          
google/cloud/firestore_v1/base_aggregation.py:243: error: "QueryResultsList" expects no type arguments, but 1 given  [type-arg]
google/cloud/firestore_v1/base_aggregation.py:243: error: "Awaitable" expects 1 type argument, but 3 given  [type-arg]
google/cloud/firestore_v1/base_aggregation.py:283: error: "AsyncStreamGenerator" expects no type arguments, but 2 given  [type-arg]
```

```
tmp/python-firestore [query-profiling-3●] » mypy -p google.cloud.firestore_v1.collection                                                                                                                                                  
google/cloud/firestore_v1/collection.py:170: error: Return type "QueryResultsList" of "get" incompatible with return type "Generator[DocumentSnapshot, Any, Any] | AsyncGenerator[DocumentSnapshot, Any]" in supertype "BaseCollectionReference"  [override]
google/cloud/firestore_v1/collection.py:177: error: "QueryResultsList" expects no type arguments, but 1 given  [type-arg]
```

```
tmp/python-firestore [query-profiling-3●] » mypy -p google.cloud.firestore_v1.query_results
google/cloud/firestore_v1/query_results.py:71: error: Incompatible return value type (got "ExplainMetrics | None", expected "ExplainMetrics")  [return-value]
```

```
tmp/python-firestore [query-profiling-3●] » mypy -p google.cloud.firestore_v1.stream_generator                                                                                                                                         
google/cloud/firestore_v1/stream_generator.py:68: error: Signature of "throw" incompatible with supertype "Generator"  [override]
google/cloud/firestore_v1/stream_generator.py:68: note:      Superclass:
google/cloud/firestore_v1/stream_generator.py:68: note:          @overload
google/cloud/firestore_v1/stream_generator.py:68: note:          def throw(self, type[BaseException], object = ..., TracebackType | None = ..., /) -> T
google/cloud/firestore_v1/stream_generator.py:68: note:          @overload
google/cloud/firestore_v1/stream_generator.py:68: note:          def throw(self, BaseException, None = ..., TracebackType | None = ..., /) -> T
google/cloud/firestore_v1/stream_generator.py:68: note:      Subclass:
google/cloud/firestore_v1/stream_generator.py:68: note:          def throw(self, exp: Any = ...) -> Any

```

Linchin:
```
google/cloud/firestore_v1/base_aggregation.py:243: error: "QueryResultsList" expects no type arguments, but 1 given  [type-arg]
```
I clean up `QueryResultsList` related type hints.

```
google/cloud/firestore_v1/base_aggregation.py:243: error: "Awaitable" expects 1 type argument, but 3 given  [type-arg]
```
Now that I think about it, I don't think get() returns a coroutine here, so I changed it to the list.

```
google/cloud/firestore_v1/base_aggregation.py:283: error: "AsyncStreamGenerator" expects no type arguments, but 2 given  [type-arg]
```
Added generic yield type for AsyncStreamGenerator, but we will be using it as `AsyncStreamGenerator[YieldType]`, instead of `AsyncStreamGenerator[YieldType, SendType]`. I'm not sure if it's worth the effort to create a second generic type just for this.

```
tmp/python-firestore [query-profiling-3●] » mypy -p google.cloud.firestore_v1.collection                                                                                                                                                  
google/cloud/firestore_v1/collection.py:170: error: Return type "QueryResultsList" of "get" incompatible with return type "Generator[DocumentSnapshot, Any, Any] | AsyncGenerator[DocumentSnapshot, Any]" in supertype "BaseCollectionReference"  [override]
google/cloud/firestore_v1/collection.py:177: error: "QueryResultsList" expects no type arguments, but 1 given  [type-arg]
```
Cleaned this up.

```
tmp/python-firestore [query-profiling-3●] » mypy -p google.cloud.firestore_v1.query_results
google/cloud/firestore_v1/query_results.py:71: error: Incompatible return value type (got "ExplainMetrics | None", expected "ExplainMetrics")  [return-value]
```
I feel like this is caused by mypy being unaware of the inner logic of the function. It either returns ExplainMetrics, or raises an exception, so there's no way it returns None. I will add `# type: ignore` to this line.


```
tmp/python-firestore [query-profiling-3●] » mypy -p google.cloud.firestore_v1.stream_generator                                                                                                                                         
google/cloud/firestore_v1/stream_generator.py:68: error: Signature of "throw" incompatible with supertype "Generator"  [override]
google/cloud/firestore_v1/stream_generator.py:68: note:      Superclass:
google/cloud/firestore_v1/stream_generator.py:68: note:          @overload
google/cloud/firestore_v1/stream_generator.py:68: note:          def throw(self, type[BaseException], object = ..., TracebackType | None = ..., /) -> T
google/cloud/firestore_v1/stream_generator.py:68: note:          @overload
google/cloud/firestore_v1/stream_generator.py:68: note:          def throw(self, BaseException, None = ..., TracebackType | None = ..., /) -> T
google/cloud/firestore_v1/stream_generator.py:68: note:      Subclass:
google/cloud/firestore_v1/stream_generator.py:68: note:          def throw(self, exp: Any = ...) -> Any
```
I'm aware of the overload incompatibility here, but if we change the method signature now, we risk being backward incompatible with the version we just released 😅

google/cloud/firestore_v1/query_results.py
```diff
@@ -68,4 +69,4 @@ def get_explain_metrics(self) -> ExplainMetrics:
         if self._explain_options is None:
             raise QueryExplainError("explain_options not set on query.")
         else:
-            return self._explain_metrics
+            return self._explain_metrics  # type: ignore
```
daniel-sanche, line 72: > I feel like this is caused by mypy being unaware of the inner logic of the function. It either returns ExplainMetrics, or raises an exception, so there's no way it returns None. I will add # type: ignore to this line.

I don't think that's true. There's nothing stopping someone from passing explain_options, but not explain_metrics:
```
instance = QueryResultList([], object(), None)
instance.get_explain_metrics() == None
```

There are a couple ways we can address this. Maybe we want to ensure that if one is set, they both are, and store them together as a tuple? Or maybe we should just add a separate exception here if self._explain_metrics is None as well? I can't remember off the top of my head if it makes sense to have one but not the other here

---

Either way, I think adding `# type: ignore` should be a last resort. We can sometimes use `typing.cast` if mypy ever has trouble following the invariants. But I think in this case, it was actually a good catch in the logic

google/cloud/firestore_v1/stream_generator.py
```diff
@@ -30,7 +30,7 @@
 T = TypeVar("T")
 
 
-class StreamGenerator(Generator[T, Any, None]):
+class StreamGenerator(Generator[T, Any, Optional[ExplainMetrics]]):
```
daniel-sanche, line 33: > I'm aware of the overload incompatibility here, but if we change the method signature now, we risk being backward incompatible with the version we just released 😅

It looks like this passes the check:

```
def throw(self, exp=None, value=None, traceback=None) -> T:
    return self._generator.throw(exp, value, traceback)
```

This should be backwards compatible, since we're just adding new optional arguments. All existing code should be unaffected

---

Honestly though, I think removing the `=None` from `exp=None` would be best. exp is meant to be an exception, so passing None just raises a TypeError. Usually we should avoid potential breaking changes, but I don't think that applies in this case, where the default itself broken. Especially since this was such a recent addition. Let me know what you think

daniel-sanche (changes requested):

google/cloud/firestore_v1/base_aggregation.py
```diff
@@ -264,9 +253,8 @@ def get(
                 explain_metrics will be available on the returned generator.
 
         Returns:
-            QueryResultsList[AggregationResult] | Awaitable[Any, Any, List[AggregationResult]]:
+            (QueryResultsList[List[AggregationResult]] | List[List[AggregationResult]]):
```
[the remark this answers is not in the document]
Linchin: Sounds good, but after trying it out I think both`Coroutine[Any, Any, List[List[AggregationResult]]]` and `Awaitable[List[List[AggregationResult]]]` work? (maybe I'm missing something)

google/cloud/firestore_v1/query_results.py
Linchin: You are right, mypy did have a good point! When explain_options is set, ultimately server will return explain_metrics, and it should already been gathered before QueryResultList is initialized. To make QueryResultsList more robust, I can do the following:
1. When initializing, throw an error when `explain_options is not None and explain_metrics is None`. (maybe even the case `explain_options is None and explain_metrics is not None`?)
2. Add a check in `get_explain_metrics()` like the following in case `self._explain_metrics` got wiped out somewhere
```
    def get_explain_metrics(self) -> ExplainMetrics:
        if self._explain_options is None:
            raise QueryExplainError("explain_options not set on query.")
        elif self._explain_metrics is None:
            raise QueryExplainError("explain_metrics is empty despite explain_options is set.")
        else:
            return self._explain_metrics
```

daniel-sanche: That works, or you can require them to be passed as a tuple, which would solve the same typing problem with less checks:

```
class QueryResultsList:

    def __init__(
        self,
        docs: List,
        profile_data: tuple[ExplainOptions, ExplainMetrics] | None = None
    ):
```

google/cloud/firestore_v1/base_aggregation.py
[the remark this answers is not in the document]
daniel-sanche: This is what I see:

```
google/cloud/firestore_v1/async_aggregation.py:49: error: Signature of "get" incompatible with supertype "BaseAggregationQuery"  [override]
google/cloud/firestore_v1/async_aggregation.py:49: note:      Superclass:
google/cloud/firestore_v1/async_aggregation.py:49: note:          def get(self, transaction: Any = ..., retry: Retry | _MethodDefault | None = ..., timeout: float | None = ..., *, explain_options: ExplainOptions | None = ...) -> QueryResultsList[list[AggregationResult]] | Coroutine[Any, Any, list[AggregationResult]]
google/cloud/firestore_v1/async_aggregation.py:49: note:      Subclass:
google/cloud/firestore_v1/async_aggregation.py:49: note:          def get(self, transaction: Any = ..., retry: AsyncRetry | _MethodDefault | None = ..., timeout: float | None = ...) -> Coroutine[Any, Any, list[AggregationResult]]
```

But if you found it works for you, that's fine with me too! Either way, I think this can fall into the known-broken mypy issues to fix later

google/cloud/firestore_v1/stream_generator.py
Linchin: You are right, we shouldn't let `exp` be optional. Just to track our discussion offline, I tried to add type annotations for all the parameters here, but it seems mypy has trouble matching it with the annotations that are present in the parent class `Generator`. So we will skip the annotation here and let mypy infer it from the parent class.

daniel-sanche (approved):

google/cloud/firestore_v1/stream_generator.py
Linchin: Additionally, python 3.12 has [deprecated](https://docs.python.org/3/reference/expressions.html#generator.throw) parameters `value` and `traceback` in `generator.throw()`, so to be compatible with both 3.12 and older versions, we will just use `throw(*args, **kwargs)`.

google/cloud/firestore_v1/query_results.py
Linchin: done

google/cloud/firestore_v1/base_aggregation.py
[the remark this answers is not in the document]
Linchin: I see, I will revert the change and save this for later haha

PR #943, opened 2024-08-09
[DRAFT] chore: update blunderbuss.yml

daniel-sanche:
update blunderbuss to direct new issues to the right place


daniel-sanche:
It looks like owlbot is overwriting it

By default, it seems to follow the CODEOWNERS file. But we may want issues assigned to a different set than PR approvers

PR #960, opened 2024-08-21
feat: support returning computed distance and set distance thresholds on VectorQueries

NickChittle:


google/cloud/firestore_v1/async_query.py
```diff
@@ -230,17 +230,23 @@ def find_nearest(
...
+            vector_field (str): An indexed vector field to search upon. Only documents which contain
                 vectors whose dimensionality match the query_vector can be returned.
-            query_vector(Vector): The query vector that we are searching on. Must be a vector of no more
+            query_vector (Vector): The query vector that we are searching on. Must be a vector of no more
                 than 2048 dimensions.
             limit (int): The number of nearest neighbors to return. Must be a positive integer of no more than 1000.
-            distance_measure(:class:`DistanceMeasure`): The Distance Measure to use.
+            distance_measure (:class:`DistanceMeasure`): The Distance Measure to use.
+            distance_result_field (Optional[str]):
+                Name of the field to output the result of the vector distance calculation
```
daniel-sanche, line 247: What does it fall back to if the value is None?

google/cloud/firestore_v1/base_vector_query.py
```diff
@@ -69,6 +71,11 @@ def _to_protobuf(self) -> query.StructuredQuery:
         else:
             raise ValueError("Invalid distance_measure")
 
+        # Coerce ints to floats as required by the protobuf.
+        distance_threshold_proto = None
+        if self._distance_threshold:
+            distance_threshold_proto = float(self._distance_threshold)
```
daniel-sanche, line 77: If self._distance_threshold is 0, distance_threshold_proto will be None. Is that expected?

tests/system/test_system.py
```diff
@@ -218,10 +218,40 @@ def test_vector_search_collection_with_filter(client, database):
...
+        "embedding": Vector([1.0, 2.0, 3.0]),
+        "color": "red",
+        "vector_distance": 0.0,
+    }
+    assert returned[1].to_dict() == {
+        "embedding": Vector([2.0, 2.0, 3.0]),
+        "color": "red",
+        "vector_distance": 1.0,
+    }
+
```
daniel-sanche, line 249: It might be good to have some different values of distance_threshold/distance_measure in these tests. (You could use `@pytest.mark.parametrize` to test various arguments/expected results in the same test function)

google/cloud/firestore_v1/base_collection.py
```diff
@@ -550,23 +550,34 @@ def find_nearest(
         query_vector: Vector,
         limit: int,
         distance_measure: DistanceMeasure,
+        distance_result_field: Optional[str] = None,
+        distance_threshold: Optional[Union[int, float]] = None,
```
daniel-sanche, line 554: Maybe we should add a `*` before these optional args to make them keyword-only? That buys us a bit of wiggle-room to re-order them later, if we end up adding more args in the future

(I probably would have preferred to make everything after query_vector keyword-only, but it's too late for that change)

google/cloud/firestore_v1/base_vector_query.py
```diff
@@ -45,6 +45,8 @@ def __init__(self, nested_query) -> None:
         self._query_vector: Optional[Vector] = None
         self._limit: Optional[int] = None
         self._distance_measure: Optional[DistanceMeasure] = None
+        self._distance_result_field: Optional[str] = None
+        self._distance_threshold: Optional[Union[int, float]] = None
```
daniel-sanche, line 49: nit: `float` implies `Union[int, float]` when doing type annotations: https://stackoverflow.com/questions/50928592/mypy-type-hint-unionfloat-int-is-there-a-number-type

google/cloud/firestore_v1/async_query.py
NickChittle: The returned documents just won't have the distance field on it. In the case that it is set, then If you have a document like:

{ __name__: "docname", "embedding": somevector, "color": "red" }

With a distanceResultField="vector_distance" the returned document is

{ __name__: "docname", "embedding": somevector, "color": "red", "vector_distance": 4.5}

google/cloud/firestore_v1/base_vector_query.py
NickChittle: No thanks for the catch.

google/cloud/firestore_v1/async_query.py
daniel-sanche: Ok, it might be worth mentioning that in the docstring

NickChittle: Done.

google/cloud/firestore_v1/base_collection.py
NickChittle: Done.

google/cloud/firestore_v1/base_vector_query.py
NickChittle: Done.

tests/system/test_system.py
NickChittle: This was a bit tricky because not all the distance measures have the same results, for some of them I was able to make it work, for a couple others I just duplicated the bulk of the test.

daniel-sanche (approved):

PR #961, opened 2024-08-22
feat: query profiling part 2: asynchronous

Linchin:
Note: this is a branch of the synchronized PR (#938). Will rebase once #938 is merged.


google/cloud/firestore_v1/async_aggregation.py
```diff
@@ -124,14 +153,21 @@ async def _make_stream(
 
         async for response in response_iterator:
             result = _query_response_to_result(response)
-            yield result
+            if result:
```
Linchin, line 156: Note: I added this so it doesn't yield an empty list when we have `explain_options=ExplainOptions(analyze=False)`. In this case no results are returned but we get a message containing explain metrics info.

google/cloud/firestore_v1/async_query.py
```diff
@@ -371,12 +398,18 @@ async def _make_stream(
             if snapshot is not None:
                 yield snapshot
 
+            if metrics is None and response.explain_metrics:
+                metrics = response.explain_metrics
+                yield metrics
```
daniel-sanche, line 403: this metrics variable seems unnecessary to me. Can we just do this?

```
if response.explain_metrics:
    yield response.explain_metrics
```

google/cloud/firestore_v1/async_transaction.py
```diff
@@ -182,18 +188,30 @@ async def get(
...
+            or :data:`None` if the document does not exist.
         """
         kwargs = _helpers.make_retry_timeout_kwargs(retry, timeout)
         if isinstance(ref_or_query, AsyncDocumentReference):
+            if explain_options is not None:
+                warnings.warn(
+                    "explain_options not supported in transanction with "
+                    "document references and will be ignored. To use "
+                    "explain_options, use transaction with query instead."
+                )
```
daniel-sanche, line 207: Maybe good to reference the `ref_or_query` argument name here, so they know what to change?

I'm still kind of considering whether it makes sense to warn or raise an exception here. I'm curious if you have strong thoughts one way or the other

google/cloud/firestore_v1/async_transaction.py
```diff
@@ -182,18 +188,30 @@ async def get(
...
             return await self._client.get_all(
                 [ref_or_query], transaction=self, **kwargs
             )
         elif isinstance(ref_or_query, AsyncQuery):
-            return await ref_or_query.stream(transaction=self, **kwargs)
+            if explain_options is not None:
+                kwargs["explain_options"] = explain_options
+            return ref_or_query.stream(transaction=self, **kwargs)
```
daniel-sanche, line 214: is the drop of `await` intentional here?

google/cloud/firestore_v1/async_vector_query.py
```diff
@@ -65,49 +74,83 @@ async def get(
...
+        This sends a ``RunQuery`` RPC and then returns a generator which
         consumes each document returned in the stream of ``RunQueryResponse``
         messages.
 
+        .. note::
+
+           The underlying stream of responses will time out after
+           the ``max_rpc_timeout_millis`` value set in the GAPIC
+           client configuration for the ``RunQuery`` API.  Snapshots
+           not consumed from the iterator before that point will be lost.
```
daniel-sanche, line 122: is there any context behind this new comment?

google/cloud/firestore_v1/async_vector_query.py
```diff
@@ -65,49 +74,83 @@ async def get(
...
+                to a system-specified value.
+            explain_options
+                (Optional[:class:`~google.cloud.firestore_v1.query_profile.ExplainOptions`]):
+                Options to enable query profiling for this query. When set,
+                explain_metrics will be available on the returned generator.
 
         Yields:
-            :class:`~google.cloud.firestore_v1.document.DocumentSnapshot`:
+            [:class:`~google.cloud.firestore_v1.base_document.DocumentSnapshot` \
+                | google.cloud.firestore_v1.types.query_profile.ExplainMetrtics]:
```
daniel-sanche, line 144: please add a comment about the ExplainMetrics yeild

daniel-sanche (changes requested):
mostly looks good, but a few small comments

google/cloud/firestore_v1/async_query.py
Linchin: This can prevent it from yielding `explain_metrics` more than twice. I think it's unlikely to happen, should we worry about it?

google/cloud/firestore_v1/async_vector_query.py
Linchin: I think I just copied from other `_make_stream()` docstrings together with the rest of the docstring, see [all occurrences](https://github.com/search?q=repo%3Agoogleapis%2Fpython-firestore%20%22max_rpc_timeout_millis%22&type=code). They seem to only appear for query and collection - does it apply to vector queries?

google/cloud/firestore_v1/async_query.py
daniel-sanche: I think that's ok. If we somehow did get multiple metrics in a stream, the consumers of this would just use the most recent value, which seems like a fine behaviour to me

google/cloud/firestore_v1/async_transaction.py
Linchin: I think it was unintentional. I'm quite unsure whether we need `await` here: `stream()` itself is not async, but it does call an async function inside, without using `await`. Does this make `stream()` an implied async function?

daniel-sanche: Yeah lookin at the code, it looks like this is calling AsyncQuery.stream, which is an async generator, and should not need to be awaited. But now I'm wondering how the code worked before... Was this a bug? Or am I missing something?

google/cloud/firestore_v1/async_vector_query.py
daniel-sanche: Ok, I don't really think it applies to async code, but it's an internal method so I guess I could take it or leave it for now. I can look closer at this comment when we refactor this later

google/cloud/firestore_v1/async_aggregation.py
```diff
@@ -105,15 +125,22 @@ async def _make_stream(
...
+            explain_options
+                (Optional[:class:`~google.cloud.firestore_v1.query_profile.ExplainOptions`]):
+                Options to enable query profiling for this query. When set,
+                explain_metrics will be available on the returned generator.
 
         Yields:
-            :class:`~google.cloud.firestore_v1.base_aggregation.AggregationResult`:
-            The result of aggregations of this query
+            List[AggregationResult] | query_profile_pb.ExplainMetrics:
+            The result of aggregations of this query.
```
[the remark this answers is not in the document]
Linchin: Good idea! Added.

google/cloud/firestore_v1/async_transaction.py
Linchin: Originally I only added a warning because I didn't want to block people's workflows. Now that I contemplate it, if we raise an exception, it's a little less convenient for the user but perhaps it's less error prone? For example if someone sets`explain_options(analyze=False)` for document references, expecting the query to not run (thus not spending money) but then they receive a warning and the query is run. Or in some situations warnings are completely filtered out so people never see it. Maybe we should raise exception to prevent situations like this. WDYT?

daniel-sanche: I think we don't need to worry about breaking workflows here, since explain_options is a new flag

I'm a bit split on it, but I think I'd lean towards using an exception here, and then we can always loosen it to a warning if we get feedback that it makes certain things difficult.

You should add a mention of this limitation in the `explain_options` section of the docstring (and maybe `ref_or_query`?), and add a `Throws` section if you go with the exception

Linchin: Regardless I will add type hints for `ref_or_query`, and comments saying `explain_options` doesn't apply to documents.

google/cloud/firestore_v1/async_vector_query.py
Linchin: Thanks for catching this, I changed `AsyncQuery._make_stream()` too.

google/cloud/firestore_v1/async_transaction.py
Linchin: Anyway, it seemed ok to undo the change, so I undid it for now.

google/cloud/firestore_v1/async_vector_query.py
Linchin: Sounds good, I will delete it because I'm not even sure if it applies here.

google/cloud/firestore_v1/async_transaction.py
Linchin: I did the following:
- change warning to exception
- type hint for `ref_or_query`
- note restriction in `explain_options` section
- add `Raises` section

google/cloud/firestore_v1/async_query.py
Linchin: Removed all in async funcs.

google/cloud/firestore_v1/async_transaction.py
Linchin: From our discussion: 

You were right that `stream()` should not be awaited, and this was a bug the unit tests did not catch because `stream()` was overwritten by `Mock`. I made the following changes to correct it:
- remove await
- only mock the API for unit tests for `transaction.get(query)` so we can test the whole `stream()` behavior, both sync and async (unit tests of `transaction.get_all()` also need some work but I will keep it out of scope for this PR)
- add async system test to calling `transaction.get(query)` inside the transaction decorator, in addition to `query.get(transaction)`

google/cloud/firestore_v1/async_vector_query.py
Linchin: done

google/cloud/firestore_v1/async_transaction.py
daniel-sanche: thanks for adding all these extra tests, even though this was an existing bug. Very helpful!

daniel-sanche (approved):
LGTM!

PR #1013, opened 2025-01-30
feat: Added read_time as a parameter to various calls (synchronous/base classes)

### LINKED ISSUE #775: support read_time queries
The firestore API exposes a read_time option, which can be used to query the database at a specific point in time. We should expose this in the hand-written layers

gkevinzheng:
This PR adds `read_time` to API calls that map to various Firestore gRPC calls whose request objects contain `read_time` as a field.

Calls with `read_time` added:
| Client Library Call                        | gRPC Request in Call  |
| --------------------------------| --------------- |
| `AggregationQuery.get/stream`| [RunAggregationQueryRequest](https://firebase.google.com/docs/firestore/reference/rpc/google.firestore.v1#runaggregationqueryrequest) |
| `Client.get_all` | [BatchGetDocumentsRequest](https://firebase.google.com/docs/firestore/reference/rpc/google.firestore.v1#batchgetdocumentsrequest) |
| `Client.collections` | [ListCollectionIdsRequest](https://firebase.google.com/docs/firestore/reference/rpc/google.firestore.v1#listcollectionidsrequest) |
| `Collection.list_documents` | [ListDocumentsRequest](https://firebase.google.com/docs/firestore/reference/rpc/google.firestore.v1#listdocumentsrequest) |
| `Collection.get/stream` | [RunQueryRequest](https://firebase.google.com/docs/firestore/reference/rpc/google.firestore.v1#runqueryrequest) |
| `DocumentReference.get` | [BatchGetDocumentsRequest](https://firebase.google.com/docs/firestore/reference/rpc/google.firestore.v1#batchgetdocumentsrequest) |
| `DocumentReference.collections` | [ListCollectionIdsRequest](https://firebase.google.com/docs/firestore/reference/rpc/google.firestore.v1#listcollectionidsrequest) |
| `Query.get/stream` | [RunQueryRequest](https://firebase.google.com/docs/firestore/reference/rpc/google.firestore.v1#runqueryrequest) |
| `Query.list_partitions` | [PartitionQueryRequest](https://firebase.google.com/docs/firestore/reference/rpc/google.firestore.v1#partitionqueryrequest) |
| `Transaction.get_all` | [BatchGetDocumentsRequest](https://firebase.google.com/docs/firestore/reference/rpc/google.firestore.v1#batchgetdocumentsrequest) |
| `Transaction.get` | [BatchGetDocumentsRequest](https://firebase.google.com/docs/firestore/reference/rpc/google.firestore.v1#batchgetdocumentsrequest)/ [RunQueryRequest](https://firebase.google.com/docs/firestore/reference/rpc/google.firestore.v1#runqueryrequest) |

Partially fixes #775 for synchronous/base classes, async classes will be in a separate PR.


google/cloud/firestore_v1/aggregation.py
```diff
@@ -20,6 +20,7 @@
 """
 from __future__ import annotations
 
+from datetime import datetime
```
daniel-sanche, line 23: nit: just importing datetime as a module is probably better? This import overrides the "datetime" module name with the class name, which would get a bit weird if we have to use other datetime imports in the future

And it makes some of the type annotations a bit ambiguous

google/cloud/firestore_v1/aggregation.py
```diff
@@ -78,6 +80,10 @@ def get(
                 (Optional[:class:`~google.cloud.firestore_v1.query_profile.ExplainOptions`]):
                 Options to enable query profiling for this query. When set,
                 explain_metrics will be available on the returned generator.
+            read_time (Optional[datetime]): If set, reads documents as they were at the given
+                time. This must be a microsecond precision timestamp within the past one hour,
+                or if Point-in-Time Recovery is enabled, can additionally be a whole minute
+                timestamp within the past 7 days.
```
daniel-sanche, line 86: A couple questions:
- how are timezones handled? Do you expect ["aware" or "naive" datetimes?](https://docs.python.org/3/library/datetime.html#aware-and-naive-objects)
- I just did a quick search, but it seems like datetime.datetime also has microsecond precision. Do we need to mention the precision here?

google/cloud/firestore_v1/client.py
```diff
@@ -201,10 +202,12 @@ def document(self, *document_path: str) -> DocumentReference:
     def get_all(
         self,
         references: list,
-        field_paths: Iterable[str] | None = None,
-        transaction: Transaction | None = None,
+        field_paths: Iterable[str] = None,
+        transaction: Transaction = None,
```
daniel-sanche, line 206: why remove None from these? Doesn't that cause a mypy error?

google/cloud/firestore_v1/client.py
```diff
@@ -201,10 +202,12 @@ def document(self, *document_path: str) -> DocumentReference:
...
         retry: retries.Retry | object | None = gapic_v1.method.DEFAULT,
-        timeout: float | None = None,
+        timeout: float = None,
```
daniel-sanche, line 208: same comment: isn't this a type violation?

google/cloud/firestore_v1/client.py
```diff
@@ -239,13 +242,17 @@ def get_all(
                 should be retried.  Defaults to a system-specified policy.
             timeout (float): The timeout for this request.  Defaults to a
                 system-specified value.
+            read_time (Optional[datetime.datetime]): If set, reads documents as they were at the given
+                time. This must be a microsecond precision timestamp within the past one hour, or
+                if Point-in-Time Recovery is enabled, can additionally be a whole minute timestamp
+                within the past 7 days. For the most accurate results, use UTC timezone.
```
daniel-sanche, line 248: 1. I'm confused by this comment. Why "most accurate results". Wouldn't the wrong timezone give wrong results? How are timezones meant to be used?
2. Shouldn't this docstring be consistent with others? I don't see UTC note this in the previous files

tests/system/test_system.py
```diff
@@ -3206,6 +3426,52 @@ def in_transaction(transaction):
...
+            assert 2 in new_b_values
+            assert 3 in new_b_values
+
+            inner_fn_ran = True
+
+        in_transaction(transaction)
+        # make sure we didn't skip assertions in inner function
+        assert inner_fn_ran is True
+
+
```
daniel-sanche, line 3474: it would be good to test a read_time that is far in the past, or in the future, to make sure the exception is raised as you expect it to be

tests/unit/v1/test_document.py
```diff
@@ -520,7 +529,14 @@ def test_documentreference_get_with_transaction(database):
     _get_helper(use_transaction=True, database=database)
 
 
-def _collections_helper(page_size=None, retry=None, timeout=None, database=None):
+@pytest.mark.parametrize("database", [None, "somedb"])
+def test_documentreference_get_with_read_time(database):
+    _get_helper(read_time=DatetimeWithNanoseconds.now(), database=database)
```
daniel-sanche, line 534: Why use from DatetimeWithNanoseconds here instead of the regular datetime?

google/cloud/firestore_v1/client.py
gkevinzheng: Whoops, my bad. This was most likely from merging a really old branch. I'll revert it.

tests/unit/v1/test_document.py
gkevinzheng: The `_get_helper` creates a mock response which requires a `Timestamp` object for the read time, so I used `DatetimeWithNanoseconds` to get access to `read_time.timestamp_pb()` to get that field. I guess I could just use a regular `datetime` and have it be mocked accordingly.

gkevinzheng:
Closing this PR; #1050 should be the same PR without all the merge conflicts and rebase issues.

PR #1045, opened 2025-05-10
fix: update the async transactional types to not require extra awaits

### LINKED ISSUE #1044: The type of async_transactional is incorrect
The return of the decorated function evaluates correctly at runtime but the typechecker thinks that it is Coroutine[Unknown, Unknown, Unknown], which is incorrect.
#### Environment details

  - OS type and version: OSX 15.4.1
  - Python version: 3.13
  - pip version: 24.2
  - `google-cloud-firestore` version: 2.20.2

pyright 1.1.396 is used as the typechecker.

#### Steps to reproduce

  1. Use the `async_transactional` decorator as detailed in the [docs](https://cloud.google.com/firestore/docs/samples/firestore-transaction-document-update-async)
  2. The return of the decorated function evaluates correctly at runtime but the typechecker thinks that it is `Coroutine[Unknown, Unknown, Unknown]`

#### Code example

```python
import asyncio

import firebase_admin
from firebase_admin import firestore_async
from google.cloud.firestore_v1 import AsyncClient as AsyncFirestoreClient
from google.cloud.firestore_v1.async_transaction import (
    AsyncTransaction,
    async_transactional,
)


async def func(db: AsyncFirestoreClient) -> None:
    transaction = db.transaction()
    city_ref = db.collection("cities").document("SF")

    @async_transactional
    async def update_in_transaction(transaction: AsyncTransaction) -> str:
        snapshot = await city_ref.get(transaction=transaction)
        transaction.update(city_ref, {"population": snapshot.get("population") + 1})
        return "done"

    res = await update_in_transaction(transaction)
    print(type(res))  # evaluates to <class 'str'> at run time but Coroutine[Unknown, Unknown, Unknown] in the typechecker

async def func2(db: AsyncFirestoreClient) -> str:
    transaction = db.transaction()
    city_ref = db.collection("cities").document("SF")

    @async_transactional
    async def update_in_transaction(transaction: AsyncTransaction) -> str:
        snapshot = await city_ref.get(transaction=transaction)
        transaction.update(city_ref, {"population": snapshot.get("population") + 1})
        return "done"

    return await update_in_transaction(transaction)  # typechecker errors


if __name__ == "__main__":
    app = firebase_admin.initialize_app(
        options={"projectId": "poggio-local-experiments"}
    )
    db = firestore_async.client(app)
    asyncio.run(func(db=db))
    asyncio.run(func2(db=db))
```

jekozyra:
I updated the types of a few things in async_transactional to ensure that when a function wrapped with `@async_transactional` is called:
- the return type of the inner function is correctly detected as the correct final result by the typechecker
- we can pass additional arguments to the wrapped function without the typechecker complaining

Fixes https://github.com/googleapis/python-firestore/issues/1044 🦕


jekozyra:
@wu-hui could I get a review here please?

daniel-sanche (dismissed):

google/cloud/firestore_v1/async_transaction.py
```diff
@@ -41,6 +52,10 @@
     from google.cloud.firestore_v1.query_profile import ExplainOptions
 
 
+T = TypeVar("T")
+P = ParamSpec("P")
```
daniel-sanche, line 56: It looks like ParamSpec was added in python 3.10, and we need to support 3.8

jekozyra: I added `typing_extensions` to support expanded type hinting for older python versions. Let me know if that works, otherwise I'll come up with something more hacky for the signatures.

daniel-sanche: Sorry, adding an extra dependency to setup.py isn't an option unfortunately

Let me know if you're able to experiment with getting this working, or I can try to take it from here if you don't have the capacity

Thanks!

jekozyra: I'm not going to have time to continue looking at this for the forseeable future.

PR #1046, opened 2025-05-12
feat: add pipelines structure

daniel-sanche:
This is the first PR for pipelines support, adding the base structure for the new feature:
- added `PipelineSource` class to create a new pipeline pointed at a resource
  - added `client.pipeline()` to create a new PipelineSource object
- added `pipeline_expressions` file to hold set of expressions
  - Currently just holding base `Expr` and `Constant`
- added `pipeline_stages` file to hold set of stages. 
  - Currently just holding base `Stage`, `Collection`, and `GenericStage`
- added `BasePipeline`, `Pipeline`, and `AsyncPipeline` to chain together sequences of stages
  - implemented `pipleine.execute()` in sync and async classes
- added `PipelineResult` to expose resulting data back to the user

This PR contains unit tests. System tests will come later, when we have more stages


bhshkh (approved):

google/cloud/firestore_v1/pipeline_expressions.py
```diff
@@ -0,0 +1,85 @@
...
+    str,
+    int,
+    float,
+    bool,
+    datetime.datetime,
+    bytes,
+    GeoPoint,
+    Vector,
+    list,
+    Dict[str, Any],
```
MarkDuckworth, line 41: might be a candidate for `Dict[str, CONSTANT_TYPE]`

google/cloud/firestore_v1/pipeline_result.py
```diff
@@ -0,0 +1,142 @@
...
+        if not isinstance(other, PipelineResult):
+            return NotImplemented
+        return (self._ref == other._ref) and (self._fields_pb == other._fields_pb)
+
+    def data(self) -> Any:
+        """
+        Retrieves all fields in the result.
+
+        If a converter was provided to this `PipelineResult`, the result of the
+        converter's `from_firestore` method is returned.
```
MarkDuckworth, line 117: I can't tell if converters are or will be supported in Python pipelines

google/cloud/firestore_v1/pipeline_stages.py
```diff
@@ -0,0 +1,81 @@
...
+    def _pb_options(self) -> dict[str, Value]:
+        """Return optional named arguments that certain functions may support."""
+        return {}
+
+    def __repr__(self):
+        items = ("%s=%r" % (k, v) for k, v in self.__dict__.items() if k != "name")
+        return f"{self.__class__.__name__}({', '.join(items)})"
+
+
+class Collection(Stage):
```
MarkDuckworth, line 55: FWIW, in most SDKs, we tried to make adding a stage a method on Pipeline. Only in the Java SDKs we expose the Stage classes. I guess it's too early to tell how you are planning to implement the API to add stages to a Pipeline. And I guess you will do whatever is idomatic for Python.

MarkDuckworth (approved):
looks good. I put some comments for your consideration. They may not be actionable, just FYIs

google/cloud/firestore_v1/pipeline_result.py
daniel-sanche: Good catch, I ended up cutting it from this PR. I'll remove the docstring

Let me know if you think converters would be important to have in Python for launch

google/cloud/firestore_v1/pipeline_stages.py
daniel-sanche: Python currently exposes them both ways, but using the methods on pipeline is more idiomatic and user-friendly. Maybe I'll mark this class as private, and keep the pipeline methods as the official public way to add a stage

PR #1059, opened 2025-05-23
feat: Added read_time as a parameter to various calls (async classes)

### LINKED ISSUE #775: support read_time queries

gkevinzheng:
This PS is the async counterpart to #1050. It also fixes a slight system test flakiness issue with using Python generated datetime objects in one system test.

Fixes #775 for the async client.


daniel-sanche (approved):
LGTM

PR #1163, opened 2026-02-11
chore: remove 3.9 from kokoro tests

daniel-sanche:
3.9 has been removed from the kokoro base image, so we should remove the tests from our configs

3.9 is still tested in GitHub Actions


chalmerlowe (approved):
LGTM

